PR #27, opened 2021-03-07
[DNM] WIP build files for oob building in android studio

pld:
Possibly everything in this PR is bad, feel free to junk it all, I'm an Android newbie :)

This does NOT work as is, fail with an error about `sdk_version` in the `android` gradle file being undefined.

Goal is to get something that builds in Android Studio with no changes immediately on checkout.


pld:
@Andati @rkodev see notes above, my limited Android skills on display here :)

If you could take this up (or just ignore it, very aware nothing here might be useful), goal is to be able to run a build without any changes in Android Studio in checkout

rkodev (changes requested):
@pld We need to exclude all the files in the .idea and .gradle folder in the PR as well.

pld:
Cool, please go ahead and make changes to the PR

rkodev (approved):

Andati (approved):

PR #38, opened 2021-03-18
Patient registration

Andati:


pld (approved):
cool looks, for something like `patient-registration.json`, would we move that to be loaded from HAPI in the future?

Andati:
yes, questionnaire resources will be loaded from HAPI

PR #44, opened 2021-03-19
add phone, email and address to patient registration

Andati:


Andati:
TODO: Figure out how resource mapper will parse:-
{
    "linkId": "PR-telecom-value",
    "definition": "http://hl7.org/fhir/StructureDefinition/Patient#Patient.telecom.system",
    "type": "string",
    "text": "phone"
  }

Also how to hide this from displaying in UI

app/src/main/assets/patient-registration.json
```diff
@@ -62,6 +62,76 @@
...
+          ],
+          "item": [
+            {
+              "linkId": "PR-telecom-system",
+              "definition": "http://hl7.org/fhir/StructureDefinition/Patient#Patient.telecom.system",
+              "type": "string",
+              "text": "system",
+              "initial": [
+                {
+                  "valueString": "phone"
```
pld, line 87: can we force this to be numeric?

related to https://github.com/OpenSRP/fhircore/issues/46

pld (approved):

app/src/main/assets/patient-registration.json
Andati: So the function here expects a string https://github.com/hapifhir/org.hl7.fhir.core/blob/e69bad679afdc8382dd309f7341f1c43d96230dc/org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/model/ContactPoint.java#L491

Andati: It's possible to extend but let me wait for reviews here https://github.com/google/android-fhir/pull/326

pld:
Yes if you want to allow formatting and country codes can see why you'd want a string, let's merge this in but note that we'll need to investigate enforcing eg regex checks for questionnaire fields

PR #115, opened 2021-05-10
Vaccine Selection Json Created with Rendering Using Sdc and Profile View Updated as per Figma

MuhammadSalman-7214:
Vaccine Selection Json Created with Rendering Using Sdc and Profile View Updated as per Figma

### DIFF
--- a/android/gradle.properties
+++ b/android/gradle.properties
@@ -17,5 +17,3 @@ org.gradle.jvmargs=-Xmx4096m
 # org.gradle.parallel=true
 
 android.useAndroidX=true
-GITHUB_USER=MuhammadSalman-7214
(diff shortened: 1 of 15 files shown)


android/gradle.properties
```diff
@@ -17,3 +17,5 @@ org.gradle.jvmargs=-Xmx4096m
 # org.gradle.parallel=true
 
 android.useAndroidX=true
+GITHUB_USER=MuhammadSalman-7214
+GITHUB_PERSONAL_ACCESS_TOKEN=6697e09838e9df37b4b764b13d3d11a4a1ee53c9
```
Rkareko, line 21: This should not be committed

Rkareko (approved):

PR #124, opened 2021-05-13
Fix record vaccine crashes

Rkareko:


rkodev (approved):

pld (commented):
are we able to add a test that crashes without this patch but not with it?

Rkareko:
> are we able to add a test that crashes without this patch but not with it?

@pld I'll have a look into this

MuhammadSalman-7214 (approved):

PR #128, opened 2021-05-18
Registration page appears whitewashed on some devices

### LINKED ISSUE #91: Registration page appears whitewashed on some devices
When you load register new patient on the device (Pixel 4a  - Android 11 RQ2A.210405.005) the labels appear whitewashed and not visible. The issue does not appear on an emulator with similar specs. Screenshots attached

syedowaisali90:
close #91

### DIFF
--- a/android/app/src/main/res/values/styles.xml
+++ b/android/app/src/main/res/values/styles.xml
@@ -22,7 +22,7 @@
         <item name="colorPrimaryDark">@color/colorPrimaryDark</item>
         <item name="colorAccent">@color/colorAccent</item>
         <item name="android:listDivider">@color/white_slightly_opaque</item>
-
+        <item name="questionnaire_theme">@style/Theme.MaterialComponents.Light.DarkActionBar</item>
     </style>
 
     <style name="AppTheme.NoActionBar">


android/app/src/main/res/values/styles.xml
```diff
@@ -22,7 +22,7 @@
         <item name="colorPrimaryDark">@color/colorPrimaryDark</item>
         <item name="colorAccent">@color/colorAccent</item>
         <item name="android:listDivider">@color/white_slightly_opaque</item>
-
+        <item name="questionnaire_theme">@style/Theme.MaterialComponents.Light.DarkActionBar</item>
```
vincent-karuri, line 25: Does setting the `AppTheme` parent to `Theme.MaterialComponents.Light.DarkActionBar` work?

syedowaisali90: No, it won't work because the Questionnaire fragment uses its own theme but we can use a custom theme by using the `questionnaire_theme` variable.

https://github.com/google/android-fhir/blob/79c7db879726fa1492578f04edb053cf8546f9c1/datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireFragment.kt#L40-L47

vincent-karuri (approved):

PR #135, opened 2021-05-19
Sample fhir resources

f-odhiambo:
1. Added some sample resources for the ANC CQL work and also a plan definition sample for the Covax app
2. Reorganised the file and naming structure for the folders
3. The ANC CQL Sample FHIR resources may need validation by Brynn  - No units given for the different use cases. Used arbitrary values to mimic a close to real-life scenario


android/gradle.properties
```diff
@@ -16,4 +16,11 @@ org.gradle.jvmargs=-Xmx4096m
 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
 # org.gradle.parallel=true
 
-android.useAndroidX=true
\ No newline at end of file
+<<<<<<< HEAD
```
pld, line 19: looks like you have an unresolved conflict, probably need to rebase again

pld (changes requested):

pld (approved):

pld (approved):

PR #240, opened 2021-06-08
Filter patients whose second dose is overdue

### LINKED ISSUE #176: Filter patients whose second dose is overdue
Ability to use a toggle filter to query and fetch all patients who have missed their second shot and are overdue for the second dosage of the vaccine. 
Screenshot attached. This should only return results of all patient who have an  overdue status without the need to search 

Screen Shot 2021-05-28 at 12 54 21 PM

Screen Shot 2021-05-28 at 12 56 49 PM

**Some assumptions** 
1. Create a new patient on the registration page
2. Go to the profile page to record the first dose of the vaccine 
3. On record first dose, use current date as a start date when you received the 1st shot 
4. We shall use a 3-week window as the due date of the 2nd shot
5. On load patient list, if within the range before due date, show the user when you are due for the next shot, else show overdue status
6. Inpatient profile show when you were supposed to have received the 2nd shot - see screenshot. Label marked in red if overdue

Screen Shot 2021-05-28 at 1 08 16 PM

syedowaisali90:

Fixes #176 

**Type**

   Feature 

**Checklist**


android/app/src/main/java/org/smartregister/fhircore/activity/RecordVaccineActivity.kt
```diff
@@ -62,9 +77,42 @@ class RecordVaccineActivity : AppCompatActivity() {
...
+
+      val iParser: IParser = FhirContext.forR4().newJsonParser()
+      val questionnaire =
+        iParser.parseResource(
+          org.hl7.fhir.r4.model.Questionnaire::class.java,
+          viewModel.questionnaire
+        ) as
+          Questionnaire
+
+      // TODO Replace manual mapping with resource mapper
```
ndegwamartin, line 90: Looks like we have manual mapping from covaxx here. The dynamic implementation is in progress. @ekigamba could you advise?

ndegwamartin: @syedowaisali90 I'm going to add a DNM on this PR for now

android/app/src/main/java/org/smartregister/fhircore/viewmodel/PatientListViewModel.kt
```diff
@@ -81,6 +96,14 @@ class PatientListViewModel(application: Application, private val fhirEngine: Fhi
           from = (page * pageSize)
         }
 
+      val showOnlyOverdue = SharedPrefrencesHelper.read(PatientListFragment.SHOW_OVERDUE_PATIENTS)
+      searchResults = searchResults.distinctBy { it.logicalId }
+      searchResults =
+        searchResults.filter {
+          !showOnlyOverdue or
+            (fetchPatientStatus(it.logicalId).value?.status == VaccineStatus.OVERDUE)
```
vincent-karuri, line 104: `fetchPatientStatus` is a suspend function, meaning it may not return immediately. We should observer it's return value instead.

Also, I think we should only run this block if `showOnlyOverdue` is set to `true`.

Could we also rename this `showOnlyOverdue` to `shouldShowOnlyOverduePatients`.

syedowaisali90: I think we can use value because `fetchPatientStatus` is already in async operation. I have checked it's working fine

syedowaisali90: I have added if check and rename a variable

android/app/src/main/java/org/smartregister/fhircore/activity/RecordVaccineActivity.kt
pld: how far out is that work? if it's not going to happen next week, I'd rather get this merged and create an issue for this todo. So that the rest of this PR doesn't go out of date

Issue #177, opened 2021-05-29
Check CqlEvaluatorSampleApp & Integrate CQL Library as demo app into FHIR core


f-odhiambo:
This is a simple app that creates a cql-evaluator and runs the WHO ANC Recommendation A2 content. It was created using Android Studio. Link to [repo](https://github.com/DBCG/CqlEvaluatorSampleApp)

Checks 
- [x] Check the structure of the FHIR resources being loaded from the assets folder 
- [x] Test functionality of loading and execution of CQL on Android client
- [x] Check whether $measure_evaluate is being run locally and the related functions on the sample app
- [x] Document the finding on this issue/ticket
 

Sample Screenshots

<img src="https://user-images.githubusercontent.com/4540684/125922970-0d0be216-33e8-49b2-8f62-a6af107a1815.png" width="200" height="400">.  <img src="https://user-images.githubusercontent.com/4540684/125922979-b0a4332c-4d06-4645-8e61-594a7bd7db00.png" width="200" height="400">

Integration
- [x] Integrate it into FHIR core as a demo app for mADX  with some enhancements which include 
- [ ] Removed reference FHIR resources from Assets folder on the app and instead sync them from HAPI FHIR Server
- [x] List Patients from the server and not assets folder
- [ ] Add screen for Selection of patient and definition to generate the operation definition 

After this we  proceed to
- [ ] #404 
- [ ] #405 
- [ ] Run $measure_evaluate on server side and then later locally on Android client using the Workflow API

f-odhiambo:
Bug in the demo app  : Seems like there are some libraries that are not syncing anymore and closing the app to terminate before launch
1. Could not find org.opencds.cqf.cql:evaluator:1.2.0-SNAPSHOT.
2. Could not find org.opencds.cqf.cql:evaluator.builder:1.2.0-SNAPSHOT.
3. Could not find org.opencds.cqf.cql:evaluator.dagger:1.2.0-SNAPSHOT.

Bug reported here https://github.com/DBCG/CqlEvaluatorSampleApp/issues/1#issue-909546106

f-odhiambo:
Updated the libs and now working fine

rkodev:
### Summary of the Sample CQL Application

The app is able to evaluate CQL expressions written in `text/cql`. 
[...]

PR #490, opened 2021-08-13
Cql Sample Application integrated into Fhir Core with Library and Plan Definition Resources used From Server

MuhammadSalman-7214:

Fixes https://github.com/opensrp/fhircore/issues/177 

**Type**
Choose one: Feature

**Checklist**


android/app/src/main/java/org/smartregister/fhircore/activity/CqlActivity.kt
```diff
@@ -0,0 +1,194 @@
...
+    fhirTypeConverter = FhirTypeConverterFactory().create(fhirContext.version.version)
+    cqlFhirParametersConverter =
+      CqlFhirParametersConverter(fhirContext, adapterFactory, fhirTypeConverter)
+    try {
+
+      val (ancLibrary, fhirHelperLibrary) =
+        runBlocking {
+          // Switch to background thread and continue blocking UI thread.
+          withContext(Dispatchers.IO) {
+            val ancLibrary = fhirEngine.load(org.hl7.fhir.r4.model.Library::class.java, "758")
```
pld, line 91: what is that specific integer for?

android/app/src/main/java/org/smartregister/fhircore/activity/CqlActivity.kt
```diff
@@ -0,0 +1,194 @@
...
+            listOf(contentProvider),
+            CqlTranslatorOptions.defaultOptions()
+          ) {
+          // This is a hack needed to circumvent a bug that's currently present in the cql-engine.
+          // By default, the LibraryLoader checks to ensure that the same translator options are
+          // used to for all libraries,
+          // And it will re-translate if possible. Since translating CQL is not currently possible
+          // on Android (some changes to the way ModelInfos are loaded is needed) the library loader
+          // just needs to load libraries
+          // regardless of whether the options match.
```
pld, line 157: what are the situations where the options don't match?

android/app/src/main/java/org/smartregister/fhircore/activity/CqlActivity.kt
```diff
@@ -0,0 +1,194 @@
...
+          override fun translatorOptionsMatch(library: Library): Boolean {
+            return true
+            //                        EnumSet<CqlTranslator.Options> options =
```
pld, line 160: I think this can be removed instead of commented, the comment above explains why

pld (commented):
what's the medium term plan for asset storage? eg would we move the plan definition into a FHIR server?

MuhammadSalman-7214:
> what's the medium term plan for asset storage? eg would we move the plan definition into a FHIR server?

yes, we would use all resources from server

android/app/src/main/java/org/smartregister/fhircore/activity/CqlActivity.kt
MuhammadSalman-7214: This is the uploaded id of Library resource on server.

android/app/src/main/java/org/smartregister/fhircore/activity/CqlActivity.kt
MuhammadSalman-7214: haven't considered about such situations yet

android/app/src/main/java/org/smartregister/fhircore/activity/CqlActivity.kt
MuhammadSalman-7214: sure

android/app/src/main/java/org/smartregister/fhircore/activity/CqlActivity.kt
```diff
@@ -0,0 +1,187 @@
...
+    fhirTypeConverter = FhirTypeConverterFactory().create(fhirContext.version.version)
+    cqlFhirParametersConverter =
+      CqlFhirParametersConverter(fhirContext, adapterFactory, fhirTypeConverter)
+    try {
+
+      val (ancLibrary, fhirHelperLibrary) =
+        runBlocking {
+          // Switch to background thread and continue blocking UI thread.
+          withContext(Dispatchers.IO) {
+            val ancLibrary = fhirEngine.load(org.hl7.fhir.r4.model.Library::class.java, "758")
```
pld, line 91: cool, to keep track of these for later replacement, how about we add a todo

```suggestion
            // todo: load ID from configuration file or elsewhere
            val ancLibrary = fhirEngine.load(org.hl7.fhir.r4.model.Library::class.java, "758")
```

android/app/src/main/java/org/smartregister/fhircore/activity/CqlActivity.kt
pld: k so then do we need this hack? b/c if we don't have those situations I don't see why we'd need it

android/app/src/main/java/org/smartregister/fhircore/activity/CqlActivity.kt
MuhammadSalman-7214: sure

android/app/src/main/java/org/smartregister/fhircore/activity/CqlActivity.kt
MuhammadSalman-7214: yes, you are right

pld: ok cool, in that case, can we remove this section of code?

PR #492, opened 2021-08-16
Integrate the Barcode reader function to the latest merged version of SDC

### LINKED ISSUE #452: Integrate the Barcode reader function to the latest merged version of SDC
The Barcode reader is required to scan the paper card, the patient identifier that is used to track the paper card that is issued to the patient. 

The current SDC release has an issue with the barcode functionality. We need to resolve the bug(s) associated with it in the codebase to ensure that the barcode feature  works appropriately as its a critical feature for the WHO Covax SVC demo app 

N/A

owais-vd:

Fixes #452 

**Type**
Choose one: (Bug fix | Feature | Documentation | Testing | Code health | Release | Other) 

**Checklist**


pld (approved):

android/engine/src/main/java/org/smartregister/fhircore/engine/ui/questionnaire/BaseQuestionnaireFragment.kt
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+package org.smartregister.fhircore.engine.ui.questionnaire
+
+import com.google.android.fhir.datacapture.QuestionnaireFragment
+import com.google.android.fhir.datacapture.contrib.QuestionnaireItemBarCodeReaderViewHolderFactory
+
+class BaseQuestionnaireFragment : QuestionnaireFragment() {
+
+  override fun getCustomQuestionnaireItemViewHolderFactoryMatchers():
```
pld, line 24: not sure I understand this, can you explain what this is doing?

android/engine/src/main/java/org/smartregister/fhircore/engine/ui/questionnaire/QuestionnaireActivity.kt
```diff
@@ -111,21 +113,33 @@ open class QuestionnaireActivity : BaseMultiLanguageActivity(), View.OnClickList
...
+                        questionnaireViewModel.generateQuestionnaireResponse(
+                          questionnaire!!,
+                          intent
+                        )
+                      )
+                    bundleOf(
+                      Pair(BUNDLE_KEY_QUESTIONNAIRE, parser.encodeResourceToString(questionnaire)),
+                      Pair(BUNDLE_KEY_QUESTIONNAIRE_RESPONSE, parsedQuestionnaireResponse)
+                    )
+                  } catch (e: Exception) {
```
pld, line 137: can you catch something more specific here?

android/engine/src/main/java/org/smartregister/fhircore/engine/ui/questionnaire/QuestionnaireActivity.kt
```diff
@@ -138,6 +152,36 @@ open class QuestionnaireActivity : BaseMultiLanguageActivity(), View.OnClickList
...
+    run loop@{
+      forEach {
+        if (it.linkId == linkId) {
+          result = it
+          return@loop
+        } else if (it.item.isNotEmpty()) {
+          result = it.item.find(linkId, result)
+        }
+      }
+    }
```
pld, line 180: I'm not a kotlin expert, but because you don't have any side effects before the bottom return, I think you could accomplish the same result in a less complex fashion with a non-local return in the conditional:

```suggestion
    forEach {
      if (it.linkId == linkId) {
        return it
      } else if (it.item.isNotEmpty()) {
        result = it.item.find(linkId, result)
      }
    }
```

android/engine/src/main/java/org/smartregister/fhircore/engine/ui/register/BaseRegisterActivity.kt
```diff
@@ -650,4 +671,65 @@ abstract class BaseRegisterActivity :
...
+  }
+
+  fun isPatientExists(barcode: String): LiveData<Result<Boolean>> {
+    val result = MutableLiveData<Result<Boolean>>()
+
+    lifecycleScope.launch(registerViewModel.dispatcher.io()) {
+      try {
+        fhirEngine.load(Patient::class.java, barcode)
+        result.postValue(Result.success(true))
+      } catch (e: Exception) {
```
pld, line 723: can we catch something more specific?

android/engine/src/main/java/org/smartregister/fhircore/engine/ui/questionnaire/QuestionnaireActivity.kt
owais-vd: Yes, you're right that's better. Thanks

android/engine/src/main/java/org/smartregister/fhircore/engine/ui/register/BaseRegisterActivity.kt
owais-vd: done

android/engine/src/main/java/org/smartregister/fhircore/engine/ui/questionnaire/QuestionnaireActivity.kt
owais-vd: done

android/engine/src/main/java/org/smartregister/fhircore/engine/ui/questionnaire/BaseQuestionnaireFragment.kt
owais-vd: In SDC we don't have any specific type related to the barcode widget. Therefore, if we need to register any custom type in the SDC widget factory so in that case, we can register it by using the custom extension. we can add the extension in the questionnaire JSON and register in `getCustomQuestionnaireItemViewHolderFactoryMatchers()` method by filtering the extension URL.

here is an example questionnaire JSON snippet.

pld: OK I see, thank you for the explanation. Two questions:

1) are there situations where the `url` will be the barcode widget, but the `valueString` will not be `"barcode"`?
2) it looks like this function is only returning barcode widgets, should this function name be specific to that, something like the suggested? Alternatively, should the url and the value string be parameters passed to this function? With the name remaining as it is
```suggestion
  override fun getBarcodeQuestionnaireItemViewHolderFactoryMatchers():
```

owais-vd: for point 1. both `URL` and `valueString` will be required to resolve the custom widget and the value could be anything but both values i.e `URL` and `valueString` should match the questionnaire item values. you can check the below code snippet.

https://github.com/opensrp/fhircore/blob/93faf45b9e6e31311050f8a92c0ecd0c21007fa6/android/engine/src/main/java/org/smartregister/fhircore/engine/ui/questionnaire/BaseQuestionnaireFragment.kt#L29-L32

For point 2. We can't change the method name nor pass the argument because it's an overridden method and coming from the SDC library.

pld: ah right an override, so can't change signature, what I'm confused about is that the name makes this sound general, but the way we are overriding it, it's now specific to only barcode widgets.

looking forward, suppose we have a map widget we can add to a questionnaire, how would we change this class (or another) to load that widget?

owais-vd: yes we can add more widgets, as we are returning the list of `QuestionnaireItemViewHolderFactoryMatcher`. checkout this file https://github.com/google/android-fhir/blob/4c939d3796c231e65d7ef477539e92d46ecbb782/datacapturegallery/src/main/java/com/google/android/fhir/datacapture/gallery/CustomQuestionnaireFragment.kt here we are adding two custom widgets `number-picker` and `barcode`

ellykits: As @owais-vd mentioned the [QuestionnaireFragment#getCustomQuestionnaireItemViewHolderFactoryMatchers](https://github.com/google/android-fhir/blob/d3c3feda998e7c7fdc20c07b539247c892c0fb72/datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireFragment.kt#L94) function signature is from the SDK and can only be modified from there. From the code documentation, the intention was to generalize the method such that the implementer can provide a list of custom views required for their use case. So @owais-vd implementation is correct. The only alternative is to include the widgets as part of the SDK. I do not think extending the `QuestionnaireFragment` from the `engine` module is necessary, PS we conventionally name abstract classes with the prefix `Base` so as to not confuse extended classes.

android/engine/src/main/java/org/smartregister/fhircore/engine/ui/questionnaire/BaseQuestionnaireFragment.kt
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.smartregister.fhircore.engine.ui.questionnaire
+
+import com.google.android.fhir.datacapture.QuestionnaireFragment
+import com.google.android.fhir.datacapture.contrib.QuestionnaireItemBarCodeReaderViewHolderFactory
+
+class BaseQuestionnaireFragment : QuestionnaireFragment() {
```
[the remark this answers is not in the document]
owais-vd: renamed to `FhirCoreQuestionnaireFragment`

android/engine/src/main/java/org/smartregister/fhircore/engine/ui/questionnaire/BaseQuestionnaireFragment.kt
pld: ah cool, got it, so we add them to the list then filter again later on

pld (approved):

pld:
lgtm! please cover the new code and let's get it merged

owais-vd:
> lgtm! please cover the new code and let's get it merged

adding more coverage here https://github.com/opensrp/fhircore/pull/698

PR #509, opened 2021-09-01
Added class and method for evaluating CQL expressions in Fhir Engine

### LINKED ISSUE #177: Check CqlEvaluatorSampleApp & Integrate CQL Library as demo app into FHIR core

kimigxfoxy:

Fixes #177

**Type**
Choose one: (Feature) 

**Checklist**


android/engine/build.gradle
```diff
@@ -54,6 +81,28 @@ dependencies {
...
+    implementation 'com.fasterxml.jackson.core:jackson-annotations:2.12.2'
+    implementation 'org.codehaus.woodstox:woodstox-core-asl:4.4.1'
+    implementation 'javax.xml.stream:stax-api:1.0-2'
+    implementation "ca.uhn.hapi.fhir:hapi-fhir-android:5.3.0"
+    implementation 'org.opencds.cqf.cql:engine:1.5.2-SNAPSHOT'
+    implementation 'org.opencds.cqf.cql:engine.fhir:1.5.2-SNAPSHOT'
+    implementation 'org.opencds.cqf.cql:evaluator:1.2.0'
+    implementation 'org.opencds.cqf.cql:evaluator.builder:1.2.0'
+    implementation 'org.opencds.cqf.cql:evaluator.dagger:1.2.0'
+    testImplementation 'junit:junit:4.+'
```
ndegwamartin, line 102: We can remove the these testing dependencies. Fhircore for example already uses Junit 5

kimigxfoxy: @ndegwamartin addressed

android/engine/src/main/java/org/smartregister/fhircore/engine/cql/Evaluator.kt
```diff
@@ -0,0 +1,167 @@
...
+            object : TranslatingLibraryLoader(
+                ModelManager(), listOf(contentProvider),
+                CqlTranslatorOptions.defaultOptions()
+            ) {
+                // This is a hack needed to circumvent a bug that's currently present in the cql-engine.
+                // By default, the LibraryLoader checks to ensure that the same translator options are used to for all libraries,
+                // And it will re-translate if possible. Since translating CQL is not currently possible
+                // on Android (some changes to the way ModelInfos are loaded is needed) the library loader just needs to load libraries
+                // regardless of whether the options match.
+                override fun translatorOptionsMatch(library: Library): Boolean {
```
pld, line 119: When we discussed in the PR before this wasn't needed b/c we don't have a case where we are loading options that do not match, is that still true? Can we remove this?

kimigxfoxy: Thanks @pld for pointing this out. The translatorOptionsMatch is removed.

android/engine/src/main/java/org/smartregister/fhircore/engine/cql/LibraryEvaluator.kt
```diff
@@ -0,0 +1,175 @@
...
+      CqlFhirParametersConverter(fhirContext, adapterFactory, fhirTypeConverter)
+    // Load Library Content and create a LibraryContentProvider, which is the interface used by the
+    // LibraryLoader for getting library CQL/ELM/etc.
+    val libraryStream: InputStream = ByteArrayInputStream(libraryData.toByteArray())
+    val fhirHelpersStream: InputStream = ByteArrayInputStream(helperData.toByteArray())
+    val library = parser.parseResource(libraryStream)
+    val fhirHelpersLibrary = parser.parseResource(fhirHelpersStream)
+    val resources: List<IBaseResource> = Lists.newArrayList(library, fhirHelpersLibrary)
+    val bundleFactory = fhirContext.newBundleFactory()
+    val bundleLinks = BundleLinks("", null, true, BundleTypeEnum.COLLECTION)
```
ndegwamartin, line 94: Is there a way to refactor this into objects that can be created only once when the class is created?

Also do we have an opportunity to maybe refactor and move some of the helper methods to some type of Helper or Util class?

android/engine/src/main/java/org/smartregister/fhircore/engine/cql/LibraryEvaluator.kt
```diff
@@ -0,0 +1,175 @@
...
+  fun runCql(
+    libraryData: String,
+    helperData: String,
+    valueSetData: String,
+    testData: String,
+    evaluatorId: String?,
+    context: String,
+    contextLabel: String
+  ): String {
+    loadConfigs(libraryData, helperData, valueSetData, testData)
```
ndegwamartin, line 165: Do we need to load all configs again every time we invoke `runCQL` or can this be refactored for optimization?

android/engine/src/main/java/org/smartregister/fhircore/engine/cql/LibraryEvaluator.kt
```diff
@@ -0,0 +1,175 @@
...
+    val result =
+      libraryEvaluator!!.evaluate(
+        VersionedIdentifier().withId(evaluatorId),
+        Pair.of(context, contextLabel),
+        null,
+        null
+      )
+    return fhirContext.newJsonParser().encodeResourceToString(result)
```
ndegwamartin, line 173: Do we want to always return a string after running the evaluation or do we maybe return a resource and leave it up to the user to convert if they need to?

Alternatively, we can overload this method and have one that returns the _Resource result_ and one that returns the _String result._

android/engine/src/test/java/org/smartregister/fhircore/engine/cql/LibraryEvaluatorTest.kt
```diff
@@ -0,0 +1,90 @@
...
+      result = readJsonFile(ASSET_BASE_PATH + "result.json")
+    } catch (e: IOException) {
+      e.printStackTrace()
+    }
+  }
+
+  @Test
+  fun runCql() {
+    libraryEvaluator = LibraryEvaluator()
+    val auxResult =
```
ndegwamartin, line 54: It is preferable to have the _full variable name_ instead since it is self documenting

android/engine/src/test/java/org/smartregister/fhircore/engine/cql/LibraryEvaluatorTest.kt
```diff
@@ -0,0 +1,90 @@
...
+  @Before
+  fun setUp() {
+    try {
+      libraryData = readJsonFile(ASSET_BASE_PATH + "library.json")
+      helperData = readJsonFile(ASSET_BASE_PATH + "helper.json")
+      valueSetData = readJsonFile(ASSET_BASE_PATH + "valueSet.json")
+      testData = readJsonFile(ASSET_BASE_PATH + "patient.json")
+      result = readJsonFile(ASSET_BASE_PATH + "result.json")
+    } catch (e: IOException) {
+      e.printStackTrace()
```
ndegwamartin, line 47: Here we can use **Timber** instead when logging exceptions

android/engine/build.gradle
```diff
@@ -51,6 +83,23 @@ dependencies {
     implementation 'androidx.security:security-crypto:1.1.0-alpha03'
     implementation 'joda-time:joda-time:2.10.6'
     implementation 'androidx.paging:paging-runtime-ktx:3.0.1'
+    testImplementation group: 'org.mockito', name: 'mockito-core', version: '2.19.0'
```
ndegwamartin, line 86: I don't think we need this dependency as we are using `Mockk`

android/engine/src/main/java/org/smartregister/fhircore/engine/cql/LibraryEvaluator.kt
kimigxfoxy: @ndegwamartin I can move the load configs method to the class initializer. I also think Helper or Util will be an overkill as they are only used once in this class.

android/engine/src/main/java/org/smartregister/fhircore/engine/cql/LibraryEvaluator.kt
kimigxfoxy: Will do so. Thanks

android/engine/src/main/java/org/smartregister/fhircore/engine/cql/LibraryEvaluator.kt
kimigxfoxy: @ndegwamartin. Having this class only have only one responsibility is what I intended. This will allow other members to focus on other logic and not worry about parsing and loading fhir resource. The string returned is a JSON which will be used to display results. We don't intend to use the String response on other logic.  This will also help in Unit tests.

android/engine/src/test/java/org/smartregister/fhircore/engine/cql/LibraryEvaluatorTest.kt
kimigxfoxy: @ndegwamartin Ok, will work on it.

android/engine/src/test/java/org/smartregister/fhircore/engine/cql/LibraryEvaluatorTest.kt
kimigxfoxy: @ndegwamartin. Ok. will refactor

android/engine/build.gradle
kimigxfoxy: @ndegwamartin. Also will refactor

android/engine/src/main/java/org/smartregister/fhircore/engine/cql/LibraryEvaluator.kt
ndegwamartin: These calls for example might be expensive when creating the instances

```
 var parser: IParser? = null
    parser = fhirContext.newJsonParser()
    parser.setPrettyPrint(true)
    libraryVersionSelector = LibraryVersionSelector(adapterFactory)
    fhirTypeConverter = FhirTypeConverterFactory().create(fhirContext.version.version)
    cqlFhirParametersConverter =
      CqlFhirParametersConverter(fhirContext, adapterFactory, fhirTypeConverter)
```

Not sure we need to be creating them every time we invoke the `runCQL` method?

android/engine/src/main/java/org/smartregister/fhircore/engine/cql/LibraryEvaluator.kt
ndegwamartin: It is also for maintainability i.e. better legibility/for team collaboration.

android/engine/src/main/java/org/smartregister/fhircore/engine/cql/LibraryEvaluator.kt
ndegwamartin: Cool, if the CQL evaluation results will always only ever be used as as string this approach is ok. 

I'm marking this as resolved

android/engine/src/main/java/org/smartregister/fhircore/engine/cql/Evaluator.kt
kimigxfoxy: @pld Unfortunately I have to return this line, otherwise the app will crash. See comments below from author.

// **This is a hack needed to circumvent a bug that's currently present in the cql-engine**.
// By default, the LibraryLoader checks to ensure that the same translator options are used to for all libraries,
// And it will re-translate if possible. Since translating CQL is not currently possible
// on Android (some changes to the way ModelInfos are loaded is needed) the library loader just needs to load libraries
// regardless of whether the options match.

android/engine/src/main/java/org/smartregister/fhircore/engine/cql/LibraryEvaluator.kt
```diff
@@ -0,0 +1,186 @@
...
+  var fhirContext = FhirContext.forCached(FhirVersionEnum.R4)!!
+  var adapterFactory = AdapterFactory()
+  var fhirTypeConverter: FhirTypeConverter? = null
+  var cqlFhirParametersConverter: CqlFhirParametersConverter? = null
+  var libraryVersionSelector: LibraryVersionSelector? = null
+  var contentProvider: LibraryContentProvider? = null
+  var terminologyProvider: BundleTerminologyProvider? = null
+  var bundleRetrieveProvider: BundleRetrieveProvider? = null
+  var cqlEvaluator: CqlEvaluator? = null
+  var libEvaluator: LibraryEvaluator? = null
```
kimigxfoxy, line 65: @ndegwamartin done

ndegwamartin (approved):

kimigxfoxy:
Change log updated in this PR:  https://github.com/opensrp/fhircore/pull/557

PR #590, opened 2021-09-30
Quest questionnaire sync

### LINKED ISSUE #571: Quest v0.1 / G6PD: Publisher based Questionnaire sync
As a developer, I want to be able to sync and trigger Questionnaires based on a variable assigned to the logged in user and variables assigned to Questionnaires.

- Assign a `publisher` to users (eg somewhere in the user profile we list their `publisher`) and then when that user logs in we sync any Questionnaires that have the same `publisher`, so we’re using the `publisher` to define the “app”.
  - Start with a "g6pd" `publisher` and a user with "g6pd" in their profile.
- Assign `category` to describe the function of the Questionnaire
  - To start we can have a “registration” `category`, see #567 

- Sync all Questionnaires for everyone
- Use the `context` field in Questionnaire, doesn't seem as good a fit
- Hard code IDs (what we currently do)

**Additional context**
- See #567 for how we will use the `category` field

trevorgowing:
Fixes #571 

**Type**
Choose one: Feature

**Checklist**


pld:
cool, I think this needs a rebase

android/engine/src/main/java/org/smartregister/fhircore/engine/ui/login/LoginViewModel.kt
```diff
@@ -73,6 +76,20 @@ class LoginViewModel(
...
+  private fun storeUserPreferences(application: Application, responseBody: ResponseBody) {
+    val responseBodyString = responseBody.string()
+    Timber.i(responseBodyString)
+
+    val sharedPreferences = SharedPreferencesHelper.init(application.applicationContext)
+    sharedPreferences.write(USER_SHARED_PREFERENCE_KEY, responseBodyString)
+
+    val userResponse = responseBodyString.decodeJson<UserResponse>()
+
+    if (userResponse.realmAccess.roles.contains(G6PD)) {
```
ellykits, line 88: The **realm access role**, is it something we can configure via the [ApplicationConfiguration](https://github.com/opensrp/fhircore/blob/main/android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/app/ApplicationConfiguration.kt) model maybe default to `null` to applications that do not need it? The `application` can be cast to [ConfigurableApplication](https://github.com/opensrp/fhircore/blob/main/android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/app/ConfigurableApplication.kt) to access the configuration model

trevorgowing: @ellykits SharedPreference default to null. Are you thinking that it would just be clearer to explicitly default it to null?

ellykits: The `null` default is ok. I was just thinking it would be good to have the realm access role as a property of `ApplicationConfiguration` so we do not need to update the logic in future for any other application that would require the role. We can save the configured value if not `null`

trevorgowing: @ellykits Apologies I was misunderstanding you. I think that I am with you now. You want to store the roles for future use?

The full response of `GET /userInfo` is already persisted in `SharedPreferences`
https://github.com/opensrp/fhircore/blob/quest-questionnaire-sync/android/engine/src/main/java/org/smartregister/fhircore/engine/ui/login/LoginViewModel.kt#L83.

Do you think that it is worthwhile persisting the roles separately?

Also the roles are not really application configuration but rather user specific information so I am not sure that ApplicationConfiguration is the correct place for this data.

Let me know what you think.

ellykits: > Also the roles are not really application configuration but rather user specific information so I am not sure that ApplicationConfiguration is the correct place for this data.

I see. My only concern is how can we dynamically provide the realm access role. The current implementation has a conditional check for **G6PD**, I assume this would change for a different application.

ellykits: we can ahead with the current implementation but refactor this later, maybe introduce `UserConfiguration` that can be updated dynamically

ellykits (approved):

android/engine/src/main/java/org/smartregister/fhircore/engine/ui/login/LoginViewModel.kt
trevorgowing: @ellykits Okay I think that we are finally on the same page.

Yes, this is what I was trying to say in standup.

The solution we discussed was to use the roles for this but what we actually need is a key value store.

I wanted to get this in for the deadline but I do think we need to refactor it to be more dynamic.

pld (changes requested):
looks good, pls add tests

ellykits:
@pld https://github.com/opensrp/fhircore/pull/590#issuecomment-939186675 When you run the tests on Android Studio the internal jacoco logs are displayed but the actual tests run successfully. Try running jacocoTestReport for eir module for instance:

```sh
 ./gradlew :eir:jacocoTestReport -x testReleaseUnitTest --stacktrace -Dorg.gradle.jvmargs="-Xmx2048m -XX:+HeapDumpOnOutOfMemoryError"
```

Will investigate this further in a different PR. I think this should be ready for merge.

pld:
Cool merged thanks for the tip!

PR #797, opened 2021-11-25
Quest: Sync params optimization, syncing resources based on organization and publisher

### LINKED ISSUE #668: Quest v0.2 / G6PD: Optimize Quest Synchronization (Filter Resources by app id)
### Context

As per discussion https://github.com/opensrp/fhircore/discussions/676 initiated based on issue https://github.com/opensrp/fhircore/issues/642 the Quest App synchronization needs to be reordered and the resources filtered.

### Summary

The issues we are trying to resolve are minimizing unnecessary synchronization by moving synchronization to when it is applicable as well as minimizing the number of resources that are synchronized.

The ultimate goal is to improve the user experience by improving app performance and minimizing synchronization time.

### Tasks

- [ ] Add the application id as an identifiers to any create resources. e.g. `Patient`.
- [ ] Synchronize Application Configuration post login. Configuration is currently represented as a JSON object stored in the FHIR server in a Binary resource. This request requires authentication so it doesn't make sense to synchronize this resource prior to log in.
- [ ] Filter clinical resources by application id. Currently `quest-app`.
- [ ] Trigger periodic sync post log in rather than on app startup and filter resources by application id.

shoaibmushtaq25:

Fixes #668  

**Type**
Choose one: Feature

**Checklist**


android/quest/src/main/assets/sync_by_organization_publisher.json
```diff
@@ -0,0 +1,80 @@
+[
```
maimoonak, line 1: change file name to sync-config and move it to config/app folder. We would have only one sync config and eventually in composition.. so its good to have generic name

android/quest/src/main/assets/sync_by_organization_publisher.json
```diff
@@ -0,0 +1,80 @@
...
+  {
+    "resourceType": "SearchParameter",
+    "id": "Organization-identifier",
+    "url": "http://hl7.org/fhir/SearchParameter/Organization-identifier",
+    "version": "4.0.1",
+    "name": "organization",
+    "code": "organization",
```
maimoonak, line 8: can we remove extra and repeating and some of unused properties for now like url, version, id

pld (dismissed):

android/quest/src/main/java/org/smartregister/fhircore/quest/QuestApplication.kt
```diff
@@ -49,22 +57,49 @@ open class QuestApplication : Application(), ConfigurableApplication {
...
-    val publisher = getPublisher()
-    if (publisher != null) questionnaireFilterMap[Questionnaire.SP_PUBLISHER] = publisher
-    return questionnaireFilterMap
+  private fun loadSearchParams(context: Context): List<SearchParameter> {
+    val iParser: IParser = FhirContext.forR4().newJsonParser()
+    val json = context.assets.open(SYNC_CONFIG).bufferedReader().use { it.readText() }
+    val searchParameters = mutableListOf<SearchParameter>()
+
+    val jsonArrayEntry = JSONArray(json)
+    for (i in 0 until jsonArrayEntry.length()) {
```
pld, line 99: is there a way to iterate through the contents directly instead of by index?

pld (dismissed):

android/quest/src/main/java/org/smartregister/fhircore/quest/QuestApplication.kt
shoaibmushtaq25: I changed it to foreach, is it ok now?

pld: cool yea this looks like the best we can do because Kotlin doesn't have a JSON object iterator, which is what I was hoping for

pld (approved):

android/quest/src/main/java/org/smartregister/fhircore/quest/QuestApplication.kt
shoaibmushtaq25: yeah, that's true.

pld (approved):

PR #801, opened 2021-11-26
Mc 11 design client relationship links

KhumboLihonga:
Hey guys, so I added some registration functionality. Registering new clients is still giving trouble but exposed infants and contacts should be working. Please review.


ndegwamartin:
@KhumboLihonga  could you please give some more context or share a requirement specification document for the implementation of this module?

PR #839, opened 2021-12-07
Add missed auth-interceptor | Fix login flow

maimoonak:
Fixes Login flow Updates

**Type**
Choose one: Bug fix

**Checklist**

### DIFF
--- a/android/engine/src/main/java/org/smartregister/fhircore/engine/ui/login/LoginViewModel.kt
+++ b/android/engine/src/main/java/org/smartregister/fhircore/engine/ui/login/LoginViewModel.kt
@@ -26,6 +26,7 @@ import androidx.lifecycle.MutableLiveData
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import dagger.hilt.android.lifecycle.HiltViewModel
+import java.lang.RuntimeException
 import javax.inject.Inject
 import kotlinx.coroutines.launch
 import okhttp3.ResponseBody
@@ -45,7 +46,6 @@ import org.smartregister.fhircore.engine.util.extension.encodeJson
 import retrofit2.Call
 import retrofit2.Response
 import timber.log.Timber
-import java.lang.RuntimeException
 
 @HiltViewModel
 class LoginViewModel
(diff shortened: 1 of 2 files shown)


android/engine/src/main/java/org/smartregister/fhircore/engine/ui/login/LoginViewModel.kt
```diff
@@ -87,18 +97,22 @@ constructor(
...
+  // auth-resp (failure)
+  //   -> show error
+  //   -> attempt local login (true) -> goto home
+  // auth-resp (success)
+  //   -> fetch userinfo #LoginViewModel.responseBodyHandler
+  //   -> user-resp (failure)
+  //      -> show-error
+  //   -> user-resp (success)
+  //      -> store user info
+  //      -> goto home
```
ellykits, line 109: This is a public property so we can have the above comment in a detailed and formatted comment paragraph

android/engine/src/main/java/org/smartregister/fhircore/engine/ui/login/LoginViewModel.kt
```diff
@@ -59,13 +60,22 @@ constructor(
   val launchDialPad
     get() = _launchDialPad
 
+  // -> user-resp (failure)
+  //    -> show-error
+  // -> user-resp (success)
+  //    -> store user info
+  //    -> goto home
```
ellykits, line 67: This is a public property so we can have the above comment in a detailed and formatted comment paragraph

android/engine/src/main/java/org/smartregister/fhircore/engine/ui/login/LoginViewModel.kt
```diff
@@ -59,13 +60,22 @@ constructor(
...
-          _navigateToHome.value = true
-          _showProgressBar.postValue(false)
+        if (response.isSuccessful)
+          response.body()!!.run {
+            storeUserPreferences(this)
+            _showProgressBar.postValue(false)
+            _navigateToHome.value = true
+          }
+        else {
+          handleFailure(call, RuntimeException("Network call failed with $response"))
```
ellykits, line 78: Is it possible to use an exception that will not result into app crash? Unless the `RuntTimeException` is handle I think it will result into an app crash. An `IOException` or it's subclass sounds ideal

ellykits (commented):
Left some comments

android/engine/src/main/java/org/smartregister/fhircore/engine/ui/login/LoginViewModel.kt
maimoonak: It does not throw Exception. It uses handleFailure of retrofit -> rightbelow this success handler

PR #887, opened 2021-12-17
Implement language switcher on the profile page for quest and anc

ekigamba:

Fixes #[issue number] 

**Type**
Choose one: **Feature**

**Checklist**


android/engine/src/main/java/org/smartregister/fhircore/engine/ui/userprofile/UserProfileScreen.kt
```diff
@@ -130,3 +187,14 @@ fun UserProfileRow(
   }
   Divider(color = DividerColor)
 }
+
+@Composable
+@Preview(showBackground = true)
+@ExcludeFromJacocoGeneratedReport
+private fun UserProfileScreenPreview() {
+  /*val userProfileViewModel = UserProfileViewModel(SyncBroadcaster, AccountAuthenticator(), SecureSharedPreference(
```
pld, line 195: needed?

pld (dismissed):

android/engine/src/main/java/org/smartregister/fhircore/engine/ui/userprofile/UserProfileScreen.kt
ekigamba: Nope. I will remove this

ellykits (commented):
**Proposal**

I think we can move the `switchLanguages` configuration from the `RegisterViewConfiguration` to `ApplicationConfiguration`. It is an application level config and should only be set from one place for the entire application. **It may not be necessary actually** since we can count the number of languages provided in the `languages` of `ApplicationConfiguration` and update the logic in the following manner:

- **If** we have multiple languages provided in the `ApplicationConfiguration` then language switcher is activated
- **Else** inactivate the language switcher (only one language available in the `languages` config)

**NOTE**: 
1. `switchLanguages` is used in the side menu implemented `BaseRegisterActivity` (needs to be refactored)
2. The `PATIENT_REGISTER` is no longer needed in [AppConfigClassification](https://github.com/opensrp/fhircore/blob/issue-796-multi-lang-support/android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/AppConfigClassification.kt) (By default only `Login` and `Application` config classifications are common the registers can be named differently for different apps)

ellykits:
:+1: LGTM

ellykits (approved):

Issue #779, opened 2021-11-22
Quest v0.2 / G6PD : View Test Results Page


f-odhiambo:
**Is your feature request related to a problem? Please describe.**
As a G6PD app user, I would like to view the test results page of a patient post-evaluation based on a CQL expression 

**Describe the solution you'd like**
Implementation of the mock design to be able to allow the user to view  the Test results 

**Describe alternatives you've considered**
This should be done in a configurable way based on the Quest approach 

**Additional context**

<img width="299" alt="Screen Shot 2021-11-22 at 8 57 46 AM" src="https://user-images.githubusercontent.com/4540684/142808369-fa9172c3-c3c3-4e4b-8650-86b6867622e2.png">

**Acceptance Criteria** 
- [x] View G6PD and Hemoglobin level 
- [x] View G6PD status
- [ ] View recommended regimen 
- [x] On press done , go to [View Test Results Page ](https://github.com/opensrp/fhircore/issues/779)

**Minor enhancement UX** 
1. We can use Red, Amber and Green for the status Deficient, Intermediate and Normal 

Prerequisite ticket
- [ ] #778

pld:
Please describe how the configuration will be written for this so we can implement it with Quest

f-odhiambo:
Good question. On thinking through this, I think we should not even have a new screen just to show the results. We could do the CQL computation and then move to the user profile view with test results. See mock below. Will consult design team 

NB. mock not approved 

<img width="799" alt="Screen Shot 2021-11-24 at 3 34 17 PM" src="https://user-images.githubusercontent.com/4540684/143240612-bb6a1248-a080-4b7f-9c77-5dfcf2cc588b.png">





CC @rowo @HenryRae

maimoonak:
see generated data details in issue https://github.com/opensrp/fhircore/issues/781

HenryRae:
We made some updates to access test results from the profile page i.e replaced header copy with the "status and date" e.g Deficient (5-Oct-2021) see image below
<img width="304" alt="Screenshot 2022-01-05 at 13 01 59" src="https://user-images.githubusercontent.com/31955469/148198974-4899bc14-e152-43b7-8b0c-92fe8c655c81.png">


Access results after testing has not changed, thus we will keep the header copy as it is currently.
cc: @f-odhiambo @rowo

f-odhiambo:
I think we can raise a new PR to address this to have two flows 
[...]

PR #910, opened 2021-12-29
Allow running CQL with each questionnaire

maimoonak:
Fixes Part of #779 

**Type**
Choose one: (Feature | Code health

**Checklist**

### DIFF
--- a/android/engine/src/main/java/org/smartregister/fhircore/engine/cql/LibraryEvaluator.kt
+++ b/android/engine/src/main/java/org/smartregister/fhircore/engine/cql/LibraryEvaluator.kt
@@ -32,8 +32,8 @@ import org.hl7.fhir.instance.model.api.IBaseBundle
 import org.hl7.fhir.instance.model.api.IBaseResource
 import org.hl7.fhir.r4.model.Bundle
 import org.hl7.fhir.r4.model.Parameters
+import org.hl7.fhir.r4.model.Patient
 import org.hl7.fhir.r4.model.Resource
-import org.hl7.fhir.r4.model.ResourceType
 import org.json.JSONArray
 import org.json.JSONObject
 import org.opencds.cqf.cql.engine.data.CompositeDataProvider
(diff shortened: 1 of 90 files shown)


android/engine/src/main/java/org/smartregister/fhircore/engine/cql/LibraryEvaluator.kt
```diff
@@ -193,39 +197,63 @@ class LibraryEvaluator @Inject constructor() {
...
-  fun runCqlLibrary(
-    library: org.hl7.fhir.r4.model.Library,
-    helper: org.hl7.fhir.r4.model.Library,
-    valueSet: Bundle,
-    data: Bundle
+  suspend fun runCqlLibrary(
+    libraryId: String,
+    patient: Patient,
+    resources: List<Resource>,
+    // TODO refactor class by modular and single responsibility principle
```
pld, line 204: should we do this refactor now?

android/engine/src/test/java/org/smartregister/fhircore/engine/ui/questionnaire/QuestionnaireViewModelTest.kt
```diff
@@ -825,6 +828,52 @@ class QuestionnaireViewModelTest : RobolectricTest() {
     }
   }
 
+  @Test
+  fun `extractAndSaveResources() should call runCqlFor when Questionnaire uses cqf-library extenion`() {
```
pld, line 832: 👍

android/engine/src/main/java/org/smartregister/fhircore/engine/cql/LibraryEvaluator.kt
maimoonak: This would require refactor of whole class which includes the CQL for ANC as well. Since this impacting the other work we doing for quest for multiple tickets hence I did not touch the parts which can lead to massive changes. I can revisit this after finishing the current tickets as quest or this method specifically might not need massive changes rather it would be ANC module

pld (approved):

pld (approved):

Issue #955, opened 2022-01-17
Quest - Provide error handling for missing properties for configs  - Application crashes if appId is null


f-odhiambo:
**Describe the bug**
Application crashed if appId is null or not defined or specifically property appId has not been initialized

**To Reproduce**
Steps to reproduce the behaviour:

1. Go to the Quest app
2. Click on the last Questionnaire 
3. See error

**Expected behaviour**
Add error handling in case the questionnaire has null properties required for rendering/filtering 

**Screenshots**
N/A

**Smartphone (please complete the following information):**
 - Device: Emulator and potentially physical app 
 - Android version: 12
 - Version 32

**Additional context**
Log
```
2022-01-17 16:35:58.690 8377-8377/org.smartregister.fhircore.quest E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.smartregister.fhircore.quest, PID: 8377
    kotlin.UninitializedPropertyAccessException: lateinit property appId has not been initialized
        at org.smartregister.fhircore.engine.configuration.ConfigurationRegistry.getAppId(ConfigurationRegistry.kt:51)
        at org.smartregister.fhircore.quest.data.patient.PatientRepository$fetchTestForms$2.invokeSuspend(PatientRepository.kt:148)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
        at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:571)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:750)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:678)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:665)

```

f-odhiambo:
@maimoonak Do we have a PR linked to this issue ?

maimoonak:
[...]

PR #956, opened 2022-01-17
Handle experimental questionnaire - test only forms

### LINKED ISSUE #946: Quest v0.2 / G6PD :  Add Control mode workflow for G6PD
As a G6PD app user, I would like to check if the device is properly recalibrated  in either Control Mode 1 or Control Mode 2

**Description of feature**
Add both UI/UX for control mode user journey and leverage current CQL evaluation that can allow us to reuse the current CQL implementation for G6PD evaluation 
Based on a Quest approach added a Questionnaire for test calibration that is independent of the Patient/Clinical workflow.

**Control mode  table** 
Screen Shot 2022-01-13 at 10 19 06 PM

**Test Kit in normal mode**
Screen Shot 2022-01-13 at 10 23 20 PM

**Test kit in control mode** 
Screen Shot 2022-01-13 at 10 25 36 PM

More details on testing and control mode can be found here https://www.finddx.org/wp-content/uploads/2020/09/STANDARD-G6PD-test_Training_Bangladesh-FIND-icddr-Menzies_FINAL_26Nov19.pdf 

**Describe the enhancement**

1. Add a menu item to initiate the control mode test 
2. Add a Questionnaire to be able to conduct control mode, similar to the Test Questionnaire, an image of control mode and the 2 entry points for G6PD and Haemoglobin level. This should have the following fields 
> Chip No
> Device Serial number
> G6PD level
> Haemoglobin level 
3. Author and load  a CQL Library evaluation for control mode evaluation 
4. Initiate the control mode evaluation on press submit 
4. Add a test result page for Control mode level 1 or control mode level 2 
5. Terminate control mode flow once the user has calibrated the device

maimoonak:

Fixes #946 

**Type**
Choose one: Feature

**Checklist**


android/engine/src/main/java/org/smartregister/fhircore/engine/ui/questionnaire/QuestionnaireActivity.kt
```diff
@@ -137,7 +137,7 @@ open class QuestionnaireActivity : BaseMultiLanguageActivity(), View.OnClickList
           text = context.getString(R.string.done)
         } else if (editMode) {
           text = getString(R.string.edit)
-        }
+        } else if (questionnaire.experimental) text = context.getString(R.string.done)
```
pld, line 140: can something be experimental and in `editMode`? if not you can move this into the conditional on line 136

android/engine/src/main/java/org/smartregister/fhircore/engine/ui/questionnaire/QuestionnaireActivity.kt
```diff
@@ -202,19 +202,32 @@ open class QuestionnaireActivity : BaseMultiLanguageActivity(), View.OnClickList
...
+        context = this,
+        message = R.string.questionnaire_alert_test_only_message,
+        title = R.string.questionnaire_alert_test_only_title,
+        confirmButtonListener = { handleQuestionnaireSubmit() },
+        confirmButtonText = R.string.questionnaire_alert_test_only_button_title
+      )
+    else
+      showConfirmAlert(
+        context = this,
+        message = R.string.questionnaire_alert_submit_message,
```
pld, line 224: can we put these messages into meta data on the Questionnaire? then we could load these messages from the Questionnaire, if they don't exist we use a default, and avoid the branching

android/quest/src/main/java/org/smartregister/fhircore/quest/data/patient/PatientRepository.kt
```diff
@@ -149,7 +149,8 @@ constructor(
 
       result.map {
         QuestionnaireConfig(
-          appId = configurationRegistry.appId,
+          appId =
+            if (configurationRegistry.isAppIdInitialized()) configurationRegistry.appId else "",
```
pld, line 153: can we set to null instead of empty string if uninitialized? and update QuestionnaireConfig to handle that
```suggestion
            configurationRegistry.isAppIdInitialized() and configurationRegistry.appId,
```

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/patient/register/PatientRegisterActivity.kt
```diff
@@ -48,17 +54,26 @@ class PatientRegisterActivity : BaseRegisterActivity() {
...
-      NavigationMenuOption(
-        id = R.id.menu_item_clients,
-        title = getString(R.string.menu_clients),
-        iconResource = ContextCompat.getDrawable(this, R.drawable.ic_users)!!
-      ),
-      NavigationMenuOption(
-        id = R.id.menu_item_settings,
-        title = getString(R.string.menu_settings),
-        iconResource = ContextCompat.getDrawable(this, R.drawable.ic_settings)!!
+        NavigationMenuOption(
```
pld, line 57: cool this looks good, what's the strategy to make this configurable? could load a list of menu items from a config then loop through theme here adding `NavigationMenuOption`s to the list

android/quest/src/main/res/values/strings.xml
```diff
@@ -7,11 +7,13 @@
     <string name="run_cql">Run CQL</string>
     <string name="run_cql_log">CQL FULL LOG</string>
     <string name="run_cql_output">CQL OUTPUT DATA</string>
-  
+    <string name="run_control_test">Run Control Test</string>
```
pld, line 10: This certainly and line 16 need to move to a loaded config file

pld:
one move comment, I don't know if `experimental` gets across the right message, it's not obvious that something with that attribute would not save data. We could use a term more explicitly related to not saving data, like `transient`?

android/quest/src/main/res/values/strings.xml
maimoonak: done

f-odhiambo:
This PR will be addressing issues  in ticket #946

android/engine/src/main/java/org/smartregister/fhircore/engine/ui/questionnaire/QuestionnaireActivity.kt
maimoonak: If we add this to questionnaire each questionnaire would have to explicitly add this OR default (Save data) would be used. Here we want to force message 'Data will NOT be saved' incase questionnaire is experimental/test only. We would still need a check that handles separate messages for normal/test-only flow. Keeping it in string.xml also helps to use language specific messages. We can create a ticket to move these to configs in future when we have dealt with internationalization in configs

android/quest/src/main/java/org/smartregister/fhircore/quest/data/patient/PatientRepository.kt
maimoonak: Remove this from here and would handle this in #955

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/patient/register/PatientRegisterActivity.kt
maimoonak: Line 62 does this where it loads nav from config.. however, first two navigation options are for settings/register list which is not dynamic and opens a specific activity (Settings screen/Client Register). This is default app behaviour and we have not devised any strategy yet to make this configurable

maimoonak:
This experimental is a field in questionnaire https://hl7.org/fhir/questionnaire-definitions.html#Questionnaire.experimental. If we want to change it we would have to introduce separate flow specific to this feature.

android/quest/src/main/java/org/smartregister/fhircore/quest/data/patient/PatientRepository.kt
pld: are you gong to revert the change here?

android/engine/src/main/java/org/smartregister/fhircore/engine/ui/questionnaire/QuestionnaireActivity.kt
pld: OK that makes sense b/c of i18n

pld: > This experimental is a field in questionnaire https://hl7.org/fhir/questionnaire-definitions.html#Questionnaire.experimental. If we want to change it we would have to introduce separate flow specific to this feature.

Oh sorry, I didn't realize that, very good, this looks great then

pld:
> This experimental is a field in questionnaire https://hl7.org/fhir/questionnaire-definitions.html#Questionnaire.experimental. If we want to change it we would have to introduce separate flow specific to this feature.

Oh sorry, I didn't realize that, very good, this looks great then

pld (dismissed):

pld (approved):

PR #957, opened 2022-01-18
Add test case for DateUtils

FikriMilano:

Fixes #CHT

**Type**
Choose one: Testing

**Checklist**


android/engine/src/test/java/org/smartregister/fhircore/engine/util/DateUtilsTest.kt
```diff
@@ -38,6 +39,27 @@ class DateUtilsTest {
     Assert.assertEquals("3-22-2020", finalDate)
   }
 
+  @Test
+  fun `addDays() should return same date when given initialDate and pattern`() {
```
ndegwamartin, line 43: Could we change this to something to reflect that zero days were added? e.g.

`addDays should return same date when default 0 days are added`

android/engine/src/test/java/org/smartregister/fhircore/engine/util/DateUtilsTest.kt
```diff
@@ -38,6 +39,27 @@ class DateUtilsTest {
...
+  @Test
+  fun `hasPastDays should return true when given initialDate`() {
+    val initialDate = DateTimeType("2022-01-11T15:30:10.222")
+    val hasPastDays = DateUtils.hasPastDays(initialDate)
+    Assert.assertTrue(hasPastDays)
+  }
+
   @Test
   fun `String#getDate() should return correct date when given string format expected`() {
     Assert.assertEquals("2020-03-10".getDate("yyyy-MM-dd").makeItReadable(), "10-Mar-2020")
```
ndegwamartin, line 65: The **Expected** value should always be the first parameter. Please switch these around

android/engine/src/test/java/org/smartregister/fhircore/engine/util/DateUtilsTest.kt
```diff
@@ -38,6 +39,27 @@ class DateUtilsTest {
...
+
+  @Test
+  fun `hasPastDays should return true when given initialDate and days`() {
+    val initialDate = DateTimeType("2022-01-11T15:30:10.222")
+    val hasPastDays = DateUtils.hasPastDays(initialDate, -3)
+    Assert.assertTrue(hasPastDays)
+  }
+
+  @Test
+  fun `hasPastDays should return true when given initialDate`() {
```
ndegwamartin, line 57: We need to re-word these test case name as well, it is hard to tell exactly which conditions are under test

android/engine/src/test/java/org/smartregister/fhircore/engine/util/DateUtilsTest.kt
```diff
@@ -38,6 +39,27 @@ class DateUtilsTest {
...
 
+  @Test
+  fun `addDays() should return same date when given initialDate and pattern`() {
+    val finalDate =
+      DateUtils.addDays("2020-03-10 01:23:00 AM", dateTimeFormat = "yyyy-MM-dd h:mm:ss a")
+    Assert.assertEquals("3-10-2020", finalDate)
+  }
+
+  @Test
+  fun `hasPastDays should return true when given initialDate and days`() {
```
ndegwamartin, line 50: Can we re-word this to capture conditions under test and expected output?

android/engine/src/test/java/org/smartregister/fhircore/engine/util/DateUtilsTest.kt
FikriMilano: Yeah sure

android/engine/src/test/java/org/smartregister/fhircore/engine/util/DateUtilsTest.kt
FikriMilano: Okay, thanks, I missed that one

android/engine/src/test/java/org/smartregister/fhircore/engine/util/DateUtilsTest.kt
FikriMilano: Is this what you have in mind?
`hasPastDays() should return true when given valid initialDate and has negative days`

android/engine/src/test/java/org/smartregister/fhircore/engine/util/DateUtilsTest.kt
FikriMilano: How about this?
`hasPastDays() should return true when given valid initialDate and has default 0 days`

ndegwamartin: Yeah that can work

android/engine/src/test/java/org/smartregister/fhircore/engine/util/DateUtilsTest.kt
ndegwamartin: maybe something like `hasPastDays() should return true when given  date is 3 days in the past`

android/engine/src/test/java/org/smartregister/fhircore/engine/util/DateUtilsTest.kt
ndegwamartin: By looking at the test case name, it should be easy to tell what function the method is doing (and condition we are testing) without looking at the code

FikriMilano: Oh I see, yeah, thanks, I will keep that in mind

android/engine/src/test/java/org/smartregister/fhircore/engine/util/DateUtilsTest.kt
FikriMilano: Yeah, that makes sense

pld (approved):

ndegwamartin (approved):

Issue #812, opened 2021-12-01
Quest  - Sync configs via Composition resources


f-odhiambo:
**Is your feature request related to a problem? Please describe.**

1. Create Binary resource for sync-config same as assets file
2. Add the Binary to Composition as section as per template shared by Ben
3. Move the sync-filter loading to Composition in Quest
4. Associate Composition to the organization so that it can be linked with the user organization

Handle edge cases
**Test**
- [ ] (optional- check if we can use SearchParameterResource reference directly into Composition instead of using Binary)

**Additional context**
The composition would be the same as what was shared on the ticket https://github.com/opensrp/fhircore/discussions/594#discussioncomment-1635591 with an additional 'section' for sync-config.json. This would have a Binary resource reference which with content the same as  sync-config.json in assets. (We would see if we can use existing FHIR Resource (SearchParameter) instead of Binary but that would be assessed after testing Composition)

f-odhiambo:
@ellykits Can you share the list of configs we would like to migrate to the composition resource to be DONE via this task ?

pld:
I'd like to block this until we've discussed https://github.com/opensrp/fhircore/discussions/594#discussioncomment-1724738

f-odhiambo:
Implementation for this has been paused 
CC @shoaibmushtaq25 @pld

pld:
@dubdabasoduba assigning to you to pull relevant content from #594 and close this discussion, then reassign to an engineer for imlpementation

pld:
@dubdabasoduba can you get this completed by end of next week?

dubdabasoduba:
@f-odhiambo the Parameters version of the resources are added [here](https://github.com/opensrp/fhircore/issues/821#issuecomment-999422483). 

The composition resources tagging this is added [here](https://github.com/opensrp/fhircore/issues/821#issuecomment-999422483).

f-odhiambo:
@maimoonak @ellykits Let's use issue #809 to complete this tasks here . 

Closing this for now

PR #969, opened 2022-01-19
Add medication request and relevant CQL

### LINKED ISSUE #945: Quest v0.2 / G6PD : Determine recommended malaria treatment based on patient classification and add a Medication/MedicationRequest
As a G6PD app user, I would like to view the recommended malaria treatment for a patient post-evaluation based on a CQL expression

- Please refer to the image below for threshold values for recommended treatment
- Recommended treatment logic against patient classification is to be stored within the CQL lib. 
- Anyone with 16 yrs and above to receive the regimen and not lower.
- If haemoglobin reading is under or equal to 8 g/dl, PQ is not given
- Implementation of the mock design (regimen part on the test results page) to allow the user to view recommended treatment

- Edit the current CQL expression to be able to define Medication and MedicationRequest for the patient and generate the appropriate resources 

**Additional context**

**Acceptance Criteria**
- [ ] View recommended regimen as defined in the CQL statement 

**Linked Issues**
1. #779
2. #781

maimoonak:
Fixes #945 

**Type**
Choose one: Feature

**Checklist**


android/engine/src/main/java/org/smartregister/fhircore/engine/cql/R4FhirModelResolverExt.kt
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+package org.smartregister.fhircore.engine.cql
+
+import java.lang.System.out
+import org.opencds.cqf.cql.engine.fhir.model.R4FhirModelResolver
+
+class R4FhirModelResolverExt : R4FhirModelResolver() {
+  override fun resolveType(typeName: String): Class<*> {
+    // FHIR has a bug that does not allow processing inner static classes
```
pld, line 24: do you know what library this error is in? Have we created an issue or otherwise communicated the error to the community? Can you please link to reference here, thank you!

android/quest/src/main/assets/configurations/app/sync_config.json
```diff
@@ -61,32 +61,28 @@
...
-    "name": "publisher",
-    "code": "publisher",
+    "name": "_id",
+    "code": "_id",
     "base": [
       "Library"
     ],
     "type": "token",
-    "expression": ""
+    "expression": "1753,46841,46831,46823"
```
pld, line 78: what are these for?

android/quest/src/main/assets/configurations/view/detail_view_configurations.json
```diff
@@ -107,6 +111,56 @@
             }
           }
         ]
+      },
+      {
+        "filters": [
```
pld, line 116: this and below look too specific to not be loaded via a dynamic config

android/quest/src/main/resources/cql/pq-medication/cql.txt
```diff
@@ -0,0 +1,135 @@
+//Declare name and version of lib
```
pld, line 1: is there an issue to move this be loaded instead of hard-coded?

pld: same w/android/quest/src/main/resources/cql/pq-medication/data.json

pld: and I think everything in this folder

android/quest/src/main/assets/configurations/app/sync_config.json
maimoonak: These are Library ids to be loaded. This ensures that during sync it does not load libraries from server and loads only required by app. This sync config would come from server for this ticket #594

android/engine/src/main/java/org/smartregister/fhircore/engine/cql/R4FhirModelResolverExt.kt
maimoonak: https://github.com/DBCG/cql_engine/issues/537###

android/quest/src/main/assets/configurations/view/detail_view_configurations.json
maimoonak: This one load PQ regimen recommended for patient which is 8.0 weeks for deficient, and intermediate g6pd and  2.0 week for normal, and N/A for others criteria. This is determined by CQL and ots values could only be 8.0/2.0/Missing.  86859003 coded medication request know what is the value for this. dosageInstruction.timing is the data we want to show on UI. This can change to anything else into resource properties path.

Below this config we are displaying info text about why a value is shown i.e. what criteria makes it 8.0/2.0/Missing which is also determined by the value of 86859003 coded med-request. the corresponding message is picked from list based on the data we find for patient.

android/quest/src/main/resources/cql/pq-medication/cql.txt
maimoonak: resources existing into main was confusing hence I moved all resources to test folder. These resources are not hardcoded rather used by unit tests to validate if cql is generating the right output entities with given data.. This all is just for unit tests and none is used in actual code. This helps ensure that if we update CQL it produces correct results. We had a discussion to add such on server side so that it is ensured that any updates to libraries, structure map, questionnaires are tested and validated

android/quest/src/main/assets/configurations/view/detail_view_configurations.json
maimoonak: https://user-images.githubusercontent.com/4829880/150308380-9ee3f1c9-c2a9-4ac2-b902-db2e208f8498.jpg in this image 3.0, 9.0, Deficient, N/A is loaded from data and everything else is the label associated with data. The label can change according to values which have predictable values

pld: that all makes sense, my main point is that we don't want these hard-coded in the Quest app, we want to load them into it dynamically

android/quest/src/main/resources/cql/pq-medication/cql.txt
pld: cool thanks, yes putting these in the test folder makes this clear

android/quest/src/main/assets/configurations/app/sync_config.json
pld: cool sounds good, I think the issues created for this story will address this, https://github.com/opensrp/fhircore/issues/812

android/engine/src/main/java/org/smartregister/fhircore/engine/cql/R4FhirModelResolverExt.kt
pld: thanks! adding to list of issues for google sprint

android/quest/src/main/assets/configurations/view/detail_view_configurations.json
maimoonak: This can then moved to note into medication-request or text field into med-request.dosageInstructions. This intent to keep this as Label was to allow keeping it closest to presentation layer so that we can change the  text easily when ever needed.. otherwise if it is in patient data any update would require data update for all..

E.g. current text on screen is 
**'People with G6PD status Deficient are given 8 Weeks regimen'.**
Above can be changed to 
**'Patient having G6PD Deficiency are usually given 8 Weeks medicine. Make sure to follow up with your physician after 9 weeks'.**
without any changes to data rather its a change into config. otherwise older data would need update OR we would have different data for different tests...

maimoonak: These are already in configs which for now reside into assets and would be moved into Composition. So yeah with ticket mentioned this would exist on server along with all other configs

pld: OK that is sufficient, I'm certainly not suggested we mix this with patient data.

I wanted to clarify that we could move this to the loaded composition, thank you for explaining that to me. This path makes sense, we can take out the g6pd key in this config file, move it into the app defining Composition that is loaded synced on first load and then load from there, removing the g6pd specific portion of this configuration.

pld (approved):

PR #1024, opened 2022-02-03
991 | questionnaire performance

### LINKED ISSUE #991: Quest - Update QR display and edit functionality with performant option
Previously, we had to use population just to correctly map the QuestionnaireResponse and Questionnaire due to missing answers. However, this seems to have been fixed [here](https://github.com/google/android-fhir/pull/926) and we can revert to providing the QuestionnaireResponse as it is

maimoonak:
IMP: Major changes into QuestionnaireActivity which needs careful review. Tested apps eir, quest (g6pd, quest), anc and found working, code refactor in QuestionnaireActivity and relevant flow to make it readible
 
Fixes #991 

**Type**
Choose one: Code health

**Checklist**


android/quest/src/main/java/org/smartregister/fhircore/quest/QuestApplication.kt
```diff
@@ -33,6 +39,13 @@ class QuestApplication : Application(), DataCaptureConfig.Provider {
     if (BuildConfig.DEBUG) {
       Timber.plant(Timber.DebugTree())
     }
+
+    CoroutineScope(SupervisorJob() + Dispatchers.Default).launch {
```
ellykits, line 43: Any reason for using a `SupervisorJob` here? do you intend to handle the coroutine exceptions independently? I actually don't think using a `SupervisorJob` here is necessary as we only have one task.

ellykits: **SupervisorJob**

> Similar to a regular Job with the only exception that its children can fail independently of each other. A child’s failure or cancellation does not result in the supervisor’s job failing or affecting its other children, therefore a supervisor can create a unique policy for dealing with its children’s failures.

android/engine/src/main/java/org/smartregister/fhircore/engine/ui/questionnaire/QuestionnaireActivity.kt
```diff
@@ -65,26 +67,31 @@ import timber.log.Timber
...
  */
+enum class QuestionnaireType {
+  READ_ONLY,
+  EDIT,
+  DEFAULT;
+
+  fun isEditMode() = this == EDIT
+  fun isReadOnly() = this == READ_ONLY
+}
+
```
ellykits, line 78: We can move this to the same package level where we have the `QuestionnaireActivity`

android/engine/src/main/java/org/smartregister/fhircore/engine/ui/questionnaire/QuestionnaireActivity.kt
```diff
@@ -93,117 +100,117 @@ open class QuestionnaireActivity : BaseMultiLanguageActivity(), View.OnClickList
...
+    lifecycleScope.launch {
+      loadQuestionnaireAndConfig(formName)
 
-    clientIdentifier = intent.getStringExtra(QUESTIONNAIRE_ARG_PATIENT_KEY)
-    immunizationId = intent.getStringExtra(ADVERSE_EVENT_IMMUNIZATION_ITEM_KEY)
-
-    lifecycleScope.launchWhenCreated {
-      readOnly = intent.getBooleanExtra(QUESTIONNAIRE_READ_ONLY, false)
-      editMode = intent.getBooleanExtra(QUESTIONNAIRE_EDIT_MODE, false)
+      async(Dispatchers.Default) { questionnaireViewModel.libraryEvaluator.initialize() }
```
ellykits, line 117: Can we inject and use `DefaultDispatcherProvider` here so this can be replaced with `TestDispatcher` in tests.

ellykits: `async` can be replaced with `withContext`

ellykits: Is the task CPU intensive? Hence the reason for using `Default` Dispatcher?

ellykits (changes requested):

android/engine/src/main/java/org/smartregister/fhircore/engine/ui/questionnaire/QuestionnaireActivity.kt
```diff
@@ -93,117 +100,117 @@ open class QuestionnaireActivity : BaseMultiLanguageActivity(), View.OnClickList
 
   override fun onCreate(savedInstanceState: Bundle?) {
     super.onCreate(savedInstanceState)
+    requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_PORTRAIT
```
ellykits, line 103: Any reason for locking the screen orientation to portrait?

android/engine/src/main/java/org/smartregister/fhircore/engine/ui/questionnaire/QuestionnaireActivity.kt
maimoonak: Would use the withContext and DefaultDispatcherProvider. Yeah. This starts loading the FHIR and CQL classes and resources which keeps the Main thread engaged unnecessarily.

android/engine/src/main/java/org/smartregister/fhircore/engine/ui/questionnaire/QuestionnaireActivity.kt
maimoonak: The questionnaire is a long form which does not fit the landscape. Just a slight slip of hand causes screen go landscape and the activity starts again. Th reason was to block the orientation toggle if it happens in field.

android/engine/src/main/java/org/smartregister/fhircore/engine/ui/questionnaire/QuestionnaireActivity.kt
maimoonak: @ellykits [Doumentation](https://proandroiddev.com/async-operations-with-kotlin-coroutines-part-1-c51cc581ad33) says "withContext function can be used to switch the context of a coroutine and allow the execution of a block within the new context. This function suspends until the block completes and returns the result." But I actually want to unblock operation and don`t care about the outcome of the block

ellykits: Yes that is correct, as long as you provide the correct dispatcher there should be no reason to worry. Using `async` has the same effect in this context.

ellykits: If there is no need for a return value then use `launch` instead

ellykits: From the code snippet, there are 3 coroutine tasks launched in the same scope.

ellykits: You can split the tasks launch the one you want to run in parallel with the others separately. The point is, it is not recommended to use `async` when the result deferred is not used.

ellykits: My question on using the `Dispatchers.Default` as the dispatcher, from your response

>Yeah. This starts loading the FHIR and CQL classes and resources which keeps the Main thread engaged unnecessarily.

sounds like the operation is mostly I/O and is not computation intensive in that case use `dispatcher.io()` thread which is better optimized for such use cases.

maimoonak: done

ellykits (approved):

Issue #949, opened 2022-01-14
Quest - reduce time it takes to sync resources on load


f-odhiambo:
**Name of feature to enhance**
Sync optimization for the Quest application 

**Usecase/scenario** 
How can we optimize the sync process to be able to make it work more efficiently and leverage publisher fields on resources and configs on the quest app to reduce the amount of time needed to sync resources from HAPI or sync on-demand only resources that match the config parameters? 

**Description of feature**
implement a mechanism that allows the current sync operation to be as optimized as possible for optimal peek performance 

**Describe the enhancement**
- [x] Research and development  on a sync optimization strategy 
- [x] Document findings with their pros and cons 
- [x] Have a discussion to decide on the best approach  that best addresses the sync issues 
- [x] Add issues addressing how to go about the enhancements on the Sync feature

f-odhiambo:
@FikriMilano Kindly take note that the initial ticket has been reassigned due to duplication . 

Also note, If new enhancements on FHIR Engine is required document them here https://github.com/google/android-fhir/issues/1099**

FikriMilano:
@f-odhiambo so there are 2 possible ways that could be possible to optimize this:

1. Based on this Room DB benchmark, it shouldn't take too long to insert the resources into the database. So it doesn't make sense for the sync to take around 7 minutes (on my case). Therefore, perhaps we can improve the FHIR Engine, but haven't found a way to do it.
Benchmark test reference: https://medium.com/swlh/android-dbms-performance-comparison-using-the-jetpack-benchmark-library-19581bf67443
3. Have a priority initial sync, that basically only sync the necessary resource for the app to be up and running. And then do another sync in the background to fully sync the rest of the resource. But before doing a full sync, some of the features may won't work properly because of missing resources e.g. patient searching will be limited to the patients that have been sync.

maimoonak:
1 - During sync fhir-engine saves resources, and then foreach indexed/searcheable field adds an index entry based on index type into other tables. see [ResourceIndexer](https://github.com/google/android-fhir/blob/master/engine/src/main/java/com/google/android/fhir/index/ResourceIndexer.kt#L258) class. This might be leading to increased time as one resource in database can lead upto 3-7 entries into other tables.
2 - For now we do not allow app be ready or usable unless data is setup properly. @f-odhiambo can give a final verdict on this, but it depends on field usage. Shall we allow HW to roam around app without sync complete. This can lead to duplicate entries as well incase of sync in progress and app in  use by HW.
[...]

PR #1128, opened 2022-03-07
Add count param

### LINKED ISSUE #1119: [Sync Optimization] Update all API call to use a count of 100
**Description of feature**
- The current API `_count` when querying the server is 20.  This takes Approx 7mins to sync.  
- Increasing the count to 100 increases the amount of data we receive per call and reduces the sync period to approx 3 mins

**Describe the enhancement**
- [x] Update all FHIR core calls to the API to default to a 100 count. 
- [x] Make the count definition configurable, this will enable other implementations to change their count settings if need be.  This can be defined via configs.  
     - [x]  If it's not defined then the default is 100
     - [x] If defined the default becomes the new number set on the configs. 

- We can use the [application configuration](https://github.com/opensrp/fhircore/blob/main/android/anc/src/main/assets/configurations/app/application_configurations.json) to hold the count config

@pld @maimoonak What do you think about making the count definition configurable.

FikriMilano:

Fixes #1119 #949

**Type**
Choose one: Feature

**Checklist**

### DIFF
--- a/android/engine/src/main/java/org/smartregister/fhircore/engine/di/NetworkModule.kt
+++ b/android/engine/src/main/java/org/smartregister/fhircore/engine/di/NetworkModule.kt
@@ -24,6 +24,7 @@ import dagger.Module
 import dagger.Provides
 import dagger.hilt.InstallIn
 import dagger.hilt.components.SingletonComponent
+import java.util.concurrent.TimeUnit
 import okhttp3.OkHttpClient
 import okhttp3.logging.HttpLoggingInterceptor
 import org.smartregister.fhircore.engine.configuration.app.ConfigService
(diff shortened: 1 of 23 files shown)


FikriMilano:
@maimoonak  I can't run the app after merging the latest changes, even in the main branch.
The app crashes after submitting the app config name. Either Quest or ANC.

android/engine/src/main/java/org/smartregister/fhircore/engine/di/NetworkModule.kt
```diff
@@ -54,6 +55,9 @@ class NetworkModule {
     OkHttpClient.Builder()
       .addInterceptor(interceptor)
       .addInterceptor(HttpLoggingInterceptor().apply { HttpLoggingInterceptor.Level.BODY })
+      .connectTimeout(120, TimeUnit.SECONDS)
```
pld, line 58: let's move these timeouts to constants

pld (changes requested):

android/engine/src/main/java/org/smartregister/fhircore/engine/di/NetworkModule.kt
FikriMilano: Okay

ellykits (approved):

pld:
> @pld could you approve for this changes? (couldn't merge without it)

apologies for the delay, I've dismissed my stale review, you should be all set now!

PR #1153, opened 2022-03-21
APK Release configuration for ANC

### LINKED ISSUE #1152: Automate APK releases for ANC (CI/CD)
In addition to the automated APK release set up we have for Quest, we need to have a separate set up for ANC. 

ANC will be in future integrated in quest. A semi automated approach for this release is preferred where the release can be triggered manually and on demand. All other aspects of the release like building and signing the APK will still be automated by CI.

ndegwamartin:

Fixes #1152

**Checklist**

### DIFF
--- a/android/quest/build.gradle
+++ b/android/quest/build.gradle
@@ -44,7 +44,7 @@ android {
     signingConfigs {
         release {
 
-            v1SigningEnabled true
+            v1SigningEnabled false
             v2SigningEnabled true
 
             keyAlias System.getenv("KEYSTORE_ALIAS")
--- a/.github/workflows/anc-release.yml
+++ b/.github/workflows/anc-release.yml
@@ -62,8 +62,8 @@ jobs:
         with:
           repo_token: ${{ secrets.GITHUB_TOKEN }}
           file: android/anc/build/outputs/bundle/release/anc-release.aab
-          asset_name: "anc-$tag.aab"
-          tag: ${{ github.ref }}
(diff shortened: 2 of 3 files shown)


android/quest/build.gradle
```diff
@@ -44,7 +44,7 @@ android {
     signingConfigs {
         release {
 
-            v1SigningEnabled true
+            v1SigningEnabled false
```
f-odhiambo, line 47: Why is signing false? Will this be only true when we want to publish to the playstore?

.github/workflows/apk-release.yml
```diff
@@ -1,7 +1,7 @@
 # This workflow will build a Java project with Gradle
 # For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-gradle
 
-name: APK Release
+name: Quest APK Release
```
f-odhiambo, line 4: What happens to a non Quest APK release ? in the case of Malawi core?

f-odhiambo (approved):
Added a few comments , otherwise LGTM

.github/workflows/apk-release.yml
ndegwamartin: Depends if we expect that as a recurring pattern i.e. creating new modules outside of Quest. 

We can implement a solution once we add the Malawi core module

android/quest/build.gradle
ndegwamartin: v1 is false because of a security vulnerability on some Android versions. We should be okay with just using v2.

PR #1169, opened 2022-03-24
update exception handling for failed sync

### LINKED ISSUE #1148: [Quest] App toasts failed sync but still continues to sync data in  the background
**Describe the bug**
The app throws a failed sync toast on the first install but still proceed to complete and sync data 

**To Reproduce**
1. Fresh install app
2. Enter config ID: quest 
3. Add correct credentials and proceed to login 
4. Check on sync progress and after a while, a sync failure toast is given 

**Expected behaviour**
If the sync fails user should be alert on on successful retry, then the user should also be notified that sync has resumed 
message: "Sync failed but is now resumed" as an example alert/prompt  

**Screenshots/Video**
https://user-images.githubusercontent.com/4540684/159156463-a02be9c7-c97f-474c-ac13-34304878ac7e.mp4

 - Device: Google Pixel 4a
 - Android version: 12
 - Build number: [e.g. Settings -> About phone -> Build number]
 - Version : SQ1A.220205.002

**Additional context**
Might need to agree on the appropriate syn behaviour and approval of the implementation design

roywanyaga:
Toast message for failed connection, no toast message for other exceptions - only log the exception

Fixes #1148  

**Checklist**

### DIFF
--- a/android/engine/src/main/java/org/smartregister/fhircore/engine/ui/register/BaseRegisterActivity.kt
+++ b/android/engine/src/main/java/org/smartregister/fhircore/engine/ui/register/BaseRegisterActivity.kt
@@ -84,6 +84,8 @@ import org.smartregister.fhircore.engine.util.extension.showToast
 import org.smartregister.fhircore.engine.util.extension.toggleVisibility
 import retrofit2.HttpException
 import timber.log.Timber
+import java.io.InterruptedIOException
+import java.net.UnknownHostException
 
 abstract class BaseRegisterActivity :
   BaseMultiLanguageActivity(),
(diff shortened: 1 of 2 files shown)


android/engine/src/main/java/org/smartregister/fhircore/engine/ui/register/BaseRegisterActivity.kt
```diff
@@ -710,7 +712,12 @@ abstract class BaseRegisterActivity :
       showToast(getString(R.string.session_expired))
       accountAuthenticator.logout()
     } else {
-      showToast(getString(R.string.sync_failed))
+      if (exceptions.map { it.exception }.filterIsInstance<UnknownHostException>().isNotEmpty() ||
```
pld, line 715: think you can avoid looping through the exceptions list twice, by using a different closure, and you can avoid the `isNotEmpty` check by using `any`, something like

```kotlin
....any { exception: exception.isInstance(UnknownHostException) ||
                     exception.isInstance(InterruptedIOException) }
...
```

just writing that out, I think this deserves to be made a boolean utility function, like `containsAnyInstances([list of items to check], [list of classes])`

that returns true is anything in the list of items is of any of the classes in the list of classes, and false otherwise

roywanyaga: I have refactored to incorporate this. 
I'm not sure where would be ideal to place the `containsAnyInstances` function. 
cc @ellykits

roywanyaga: I have refactored to incorporate this. 
I'm not sure where would be ideal to place the `containsAnyInstances` function. 
cc @ellykits

pld (approved):
lgtm! but agree we should wait for @ellykits to advise on where to put that function before we merge

ellykits (approved):

PR #1224, opened 2022-04-24
1219 - Quest - Patient attributes and Address attributes

### LINKED ISSUE #1219: Add and extract attributes to Patient model
Currently, in the Quest app, there is a Patient model that only has the following attributes: id, identifier, name, gender, age, address, & additional data. However, for the Tingathe app, we need some additional attributes that we are saving to the server.

When looking at the Patient FHIR specification, we need the following additional attributes:

- telcom
- address
- contact
- generalPractitioner
- managingOrganization
- link

The Patient model does have an address attribute but it is extracted as a string as a result of a combination of the individual address attributes. However, we need to be able to access the individual attributes within the address itself. Specifically, we need to access:

- district
- state
- text

AbdulWahabMemon:

Fixes #1219 

**Checklist**


KhumboLihonga:
This looks great. Just a quick comment, telecom should be a List of strings as a CHW should be able to save multiple phone numbers for a patient.

ellykits:
Can we discuss how to include these attributes to the model that will be used in the single apk architecture?

AbdulWahabMemon:
> @AbdulWahabMemon This looks great. Just a quick comment, telecom should be a List of strings as a CHW should be able to save multiple phone numbers for a patient.

@KhumboLihonga patient model attribute "telecom" also updated with list of strings

AbdulWahabMemon:
@dubdabasoduba for SAA refactoring and targeting with implement-generic-app-register, there is seperate PR #1239.

android/engine/src/main/java/org/smartregister/fhircore/engine/util/extension/PatientExtension.kt
```diff
@@ -135,6 +135,36 @@ fun Patient.extractAddress(): String {
   }
 }
 
+fun Patient.extractAddressDistrict(): String {
+  if (!hasAddress()) return ""
+  return with(addressFirstRep) { this.district ?: "" }
```
pld, line 140: can we combine these 2 lines w/a ternary here and below?

```kotlin
return with(addressFirstRep) { this.district ?: "" } if (hasAddress()) else ""
```

ellykits (approved):

PR #1239, opened 2022-04-27
1219 patient item and address attributes 

### LINKED ISSUE #1219: Add and extract attributes to Patient model

AbdulWahabMemon:

Fixes #1219

**Checklist**


android/engine/src/main/java/org/smartregister/fhircore/engine/util/extension/PatientExtension.kt
```diff
@@ -135,6 +135,36 @@ fun Patient.extractAddress(): String {
...
+
+fun Patient.extractAddressText(): String {
+  if (!hasAddress()) return ""
+  return with(addressFirstRep) { this.text ?: "" }
+}
+
+fun Patient.extractTelecom(): List<String>? {
+  if (!hasTelecom()) return null
+  return telecom.map { it.value }
+}
```
ellynerdstone, line 156: It's possible to have these implementation in one function and use utility class `org.smartregister.fhircore.engine.util.fhirpath.FhirPathDataExtractor` to extract the relevant properties from the patient resource.

ellykits, line 156: It's possible to have these implementation in one function and use utility class `org.smartregister.fhircore.engine.util.fhirpath.FhirPathDataExtractor` to extract the relevant properties from the patient resource.

ellykits:
@AbdulWahabMemon I think we can close this. As you are working on the mwcore register implementation in another branch; this functionality can be added to your other branch.

AbdulWahabMemon:
> @AbdulWahabMemon I think we can close this. As you are working on the mwcore register implementation in another branch; this functionality can be added to your other branch.

The requested changes with FhirPathDataExtractor also done in another branch, so we can close this one.

PR #1296, opened 2022-05-19
Fetch and save logged in practitioner details

### LINKED ISSUE #1282: [Quest/eCBIS] Sync to server after creation of removal of family or family member fails.
**Describe the bug**
- When you create a family member or a new family and then try to sync you will notice a _Sync failed notification_ 
- The logs show that this is failing due to a malformed Bundle but I can't seem to find the Bundle in the local SQLite App DB. 

**To Reproduce**
1. Go to any family profile and add a family member
2. Click on the sync button and wait till sync is complete
3. Click sync on another device to check on the family member you added. 

**Expected behaviour**
- [ ]  Any time a change is made on any of the resources and sync is clicked, then all the local resources should be pushed to the server and the remote resources pushed to the device. 

**Screenshots**

 - Device: Emulator
 - Android version: latest eCBIS release from the **_implement-generic-app-register_** branch

**Additional context**
AppID : `ecbis-saa`
Username: `ecbis`

ellykits:
Signed-off-by: Elly Kitoto <junkmailstoelly@gmail.com>

Fixes #1282

**Checklist**

### DIFF
--- a/android/engine/src/main/java/org/smartregister/fhircore/engine/ui/login/LoginScreen.kt
+++ b/android/engine/src/main/java/org/smartregister/fhircore/engine/ui/login/LoginScreen.kt
@@ -98,7 +98,7 @@ fun LoginScreen(loginViewModel: LoginViewModel) {
   )
   val username by loginViewModel.username.observeAsState("")
   val password by loginViewModel.password.observeAsState("")
-  val loginErrorState by loginViewModel.loginErrorState.observeAsState(LoginErrorState.NO_ERROR)
+  val loginErrorState by loginViewModel.loginErrorState.observeAsState(null)
   val showProgressBar by loginViewModel.showProgressBar.observeAsState(false)
 
   LoginPage(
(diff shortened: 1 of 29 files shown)


pld (dismissed):

android/engine/src/main/java/org/smartregister/fhircore/engine/ui/login/LoginScreen.kt
```diff
@@ -251,20 +251,31 @@ fun LoginPage(
...
-                .testTag(LOGIN_ERROR_TEXT_TAG)
-          )
-        }
-        Spacer(modifier = modifier.height(40.dp))
+        Text(
+          fontSize = 14.sp,
+          color = MaterialTheme.colors.error,
+          text =
+            when (loginErrorState) {
+              LoginErrorState.NO_ERROR -> ""
```
pld, line 259: do we need the no_error state? could we handle this with an `else` for the `when`?

pld: then whenever we set to this constant we'd set to null instead

ellykits: We do not need it actually. I will make the `loginErrorState` parameter nullable.

ellykits: I have explicitly handled `null` condition in the `when` statement; we may in future need to further categorize the error messages.

pld: ok good call, we wouldn't want to bury an error if it doesn't explicitly match the when conditions

PR #1349, opened 2022-06-08
Add Group.Characteristic model support in StructureMaps extraction

ekigamba:

Fixes #[issue number] 

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Engine | Integrate Group resource for family representation
 - Engine | Performance improvement when openning questionnaires and saving QRs
 - Engine | Add debug mode to optionally load local config
+- Engine | Add support for Group.Characteristic model in StructureMap extraction
 
 ### Fixed
 - ANC | Resolved a bug on measure reporting : App crash when All patient selected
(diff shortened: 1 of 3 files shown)


pld (dismissed):
Cool, can you provide some context, why is this here and not in the sdk?

ekigamba:
It's a optional param on the SDK. I don't remember clearly, but the SDK team prefers that some of these functionalities are not built into the SDK. 

They are currently thinking of removing the StructureMap-based extraction code from the SDK as an option https://github.com/google/android-fhir/issues/1366#issuecomment-1134952147

pld:
Ok thanks, yeah if we have to maintain these types of extensions then I see the argument for why have it in the SDK at all.

The other side of the argument might be on uniformity of approach to structuremaps across implements of the SDK that use them, but we might need code like this in the SDK for that to be meaningful.

pld:
I'm also wondering if there's a way to generate those types, this feels ad hoc

ekigamba:
> I'm also wondering if there's a way to generate those types, this feels ad hoc

I believe there should be a better way to do this. We can possibly create an issue for this since the adoption is increasing and the effort for implementing the code generation for this is. However, we'll also need to provide the missing documentation for these models because the names are only found inside the JSON model schemas. On the FHIR documentation, these models are termed as `BackboneElement` only.

We can merge this in and create another issue to discuss and implement the dynamic code generation for this

cc @Rkareko

ekigamba:
> Ok thanks, yeah if we have to maintain these types of extensions then I see the argument for why have it in the SDK at all.
> 
> The other side of the argument might be on uniformity of approach to structuremaps across implements of the SDK that use them, but we might need code like this in the SDK for that to be meaningful.

We can include this as a task on the issue so that we can propose and/or discuss with the SDK team on this

cc @Rkareko

Rkareko (approved):

ekigamba:
The link to the issue is here @pld @Rkareko

PR #1421, opened 2022-07-08
Run rules against resources

Rkareko:

Fixes #[issue number] 

**Checklist**


android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
```diff
@@ -63,16 +64,32 @@ class RulesFactory @Inject constructor(val configurationRegistry: ConfigurationR
     facts.put(FHIR_PATH_DATA_EXTRACTOR, fhirPathDataExtractor)
```
FikriMilano, line 64: Curiosity question.

Is the reason for saving this `fhirPathDataExtractor` is because of the rules actions needed to have an access to that class?

Rkareko: Yes. That's correct.

PR #1709, opened 2022-09-30
Sync by patient resources

### LINKED ISSUE #1658: [Client Side | Sync Enhancement] Identify resources to sync from the server.
**Context**
- The app needs to know what patient data-related resources we need to fetch from the server. 
- We would like to automate this process. This process is to run every time the user login using the `username`/`password`
   - This is to account for changes on the config files. 

**Description of enhancement**
- [x] Fetch all the `register` and `profile` configs from the SQLite DB. These are references by the Composition Resource. We can use it to identify the Binary Resources holding the said resources. 
     - [x] Traverse each fo the `register` & `profile` configs checking the `baseResource` & `relatedResource` keys. These keys define the resources the register or profile uses. 
     - [x] Get the ResourceTypes referenced in the 2 keys mentioned above and generate a JSON array of these Resource Types.
          - [x] The string entries in the array should unique.  
     - [x] Save this JSON array in the shared preferences of the app. We can use the key `remoteSyncResources` 
- Update the sync from server functionality to only request for the ResourceTypes on the `remoteSyncResources`

**Acceptance criteria**
- [ ] Sync resources identification should be triggered after every `username`/`password` login.  
- [x] All the ResourceTypes referenced on the `register` & `profile` configs should be part of the JSON array stored on the `remoteSyncResources` key in the shared preferences. 
- [x] The Patient data sync should only target the resources in the `remoteSyncResources`

**More information on why we need to make this change**
- https://github.com/opensrp/fhircore/discussions/1612

maimoonak:
Fixes #1658 

**Checklist**


android/engine/src/main/java/org/smartregister/fhircore/engine/ui/appsetting/AppSettingViewModel.kt
```diff
@@ -98,11 +101,29 @@ constructor(
...
+              binary.data.decodeToString().decodeBase64()!!.string(Charset.defaultCharset()).let {
+                val config =
+                  it.tryDecodeJson<RegisterConfiguration>()
+                    ?: it.tryDecodeJson<ProfileConfiguration>()
+
+                when (config) {
+                  is RegisterConfiguration ->
+                    config.fhirResource.dependentResourceTypes(patientRelatedResourceTypes)
+                  is ProfileConfiguration ->
+                    config.fhirResource.dependentResourceTypes(patientRelatedResourceTypes)
```
pld, line 116: because you only need a single branch, this is better suited for an if, or at least combine the conditions

```suggestion
                when (config) {
                  is ProfileConfiguration, RegisterConfiguration ->
                    config.fhirResource.dependentResourceTypes(patientRelatedResourceTypes)
```

android/engine/src/main/java/org/smartregister/fhircore/engine/util/extension/JsonExtension.kt
```diff
@@ -31,6 +32,10 @@ val json = Json {
 inline fun <reified T> String.decodeJson(jsonInstance: Json? = null): T =
   jsonInstance?.decodeFromString(this) ?: json.decodeFromString(this)
 
+/** Decode string to an entity of type [T] */
```
pld, line 35: can you please expand the doc string here to have at least an @return and @param following the multiline format here, https://kotlinlang.org/docs/kotlin-doc.html#kdoc-syntax

(if you could also do that for the pre-existing line function on line 40, that'd be much appreciated :D)

pld (commented):
minor style, looks good!

android/engine/src/main/java/org/smartregister/fhircore/engine/ui/appsetting/AppSettingViewModel.kt
maimoonak: I would change it to if else. The reason is it does smart cast when `is` is used i.e. smart cast to RegisterConfiguration. Unfortunately the properties for both register/profile config are similar that why the code looks repeating. Ideally we should have used a parent class for both classes for same properties. 
Using comma separated values it gives compile error 'Incompatible types: ProfileConfiguration.Companion and Configuration?'

pld (dismissed):

pld (approved):

PR #1743, opened 2022-10-17
Save questionnaire responses as in-progress

### LINKED ISSUE #1631: [Questionnaires] Adding the ability to save QuestionnaireResponse as In Progress
**Describe the feature**
- Currently, when a user exits a form without saving they are only presented with the dialog letting them know that the form content will be discarded. 
Screenshot_20220914_151214 (copy)
- We however want to give the user the ability to save the  QuestionnaireResponse as `In Progress` instead of discarding the form content. 

**Proposed Technical Implementation**
- [x] Add a boolean `canSaveDraft` to the method launching the Questionnaire. 
    - [x] This boolean will be `false` by default.  
- [x] If the `canSaveDraft` is set to `true` then the dialog that appears when a user exits the form will update to have 2 options. 
    - [x] **Save Partially** -- Allows the user to save the questionnaire partial by setting the QuestionnaireResponse `QuestionnaireResponse.status` property to `in progress`, then redirects to the screen before the questionnaire render. 
    - [x] **Discard** -- This discards all the inputs on the form and closes the Questionnaire render page, then redirects to the screen before the Questionnaire.
- [x] Resource extraction should only be done when the `QuestionnaireResponse.status` == `completed`  

**Related SDK Issue**
- https://github.com/google/android-fhir/issues/1608

**Related Discussion**
- https://github.com/opensrp/fhircore/discussions/1557

allan-on:

Fixes #1631 

**Checklist**


android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/QuestionnaireConfig.kt
```diff
@@ -34,7 +34,8 @@ data class QuestionnaireConfig(
   val resourceType: String? = null,
   val confirmationDialog: ConfirmationDialog? = null,
   val groupResource: GroupResourceConfig? = null,
-  val taskId: String? = null
+  val taskId: String? = null,
+  val saveDraft: Boolean = false
```
ellykits, line 38: Do we expect this property to be dynamic? such that it can be computed via Rules engine?

allan-on: Hi @ellykits I don't expect so. I can't think of a scenario where we'll have this computed via Rules Engine. 
@dubdabasoduba Thoughts?

ellykits: Noted.

ellykits (approved):

Issue #1778, opened 2022-10-27
[Sync Strategies] Fix the resource tagging functionality


dubdabasoduba:
**Describe the bug**
- We tag resources with practitioner's details in order to facilitate the sync of data down to the applications. 
- We have a couple of bugs with this sync piece i.e
- [ ] Currently, the sync strategies are read from the `gradle` files. This means if someone wanted to change the sync strategies they would have to generate a new APK. 
- [ ] Currently, the code tags resources based on the `syncStrategy` configuration  on `application_config.json`. i.e if the config has `Location` & `Organization` then it will tag the resources with the 2. If we only have the `Location` then it will tag the resources with the `Location` only. 

**Expected behavior**
- [x] All the resources should always be tagged by all 4 tags (Practitioner, Organization, CareTeam & Location). The `syncStrategy` configuration on the `application_config.json` should only be used for sync filtering.  e.g if you have the `Location` on the syncStrategy then the filter will use the practitioner Location to sync data down to the app.
- [x] The sync strategies should always be defined on the `application_config.json`

FikriMilano:
I need time to revert it back from `build.gradle` to `application_config.json`
Will most likely write a new code

FikriMilano:
@dubdabasoduba 
I refactored the tagging so that it will always be appended, not depending on sync strategy.
If an Organization / CareTeam / Location / Practitioner ID is somehow not defined in Practitioner Details, by default it will append "Not defined"

FikriMilano:
Currently refactoring the sync strategy to go back to application config

PR #1779, opened 2022-10-27
Update configurable views and app JSON configurations

### LINKED ISSUE #1761: [Phase 1] UI/UX and functionality updates for the configuration views.
**Context**
- After the review of the configuration work we have identified a couple of places where we need to fit bugs or tweak the UI to fit the Design team's recommendation. 
- All the changes required are added to the issue here. 

**Implementation**
- [x] Update the Service Button view to the look and feel to fit the attached screenshot. Also, update the functionality to meet the explanation in the screenshot.  
     - [x] Service button greyed out without a border -- This is when the status is `UP COMING`
         - [ ] The upcoming status disable the button. If the button is clicked while in this state a SnackBar with a user-configured message should be shown. 
     - [x] Service button with a blue border and blue text -- This is when the button status is `DUE`
     - [x] Service button with red background and white text -- This is when the service button status is `OVER DUE` 
Screenshot 2022-10-19 at 14 57 35

- [x] Add a `enable` attribute for the fab button. If the fab button `enable = false` then the button should have a look and feel as the image attached below. This should also apply to the extended fab actions. 
    - [ ] If the disabled fab is clicked then a SnackBar with a user-configured message should be shown. 
diabled-fab

- [x] Add the ability to add icons to the fab button. The image below shows the position of the icon. This should behaviour show be the same for the extended fab actions. 
- [x] Adding the `weight` attribute the to row view definition.  This will help correctly adapt the UI when have to disable different UI components.  
- [x] Add a vertical spacer view for the views in a ROW view.  
- [x] Add a border-radius attribute to the different views we have i.e ROW, COLUMN, CARD, SERVICE CARD, SERVICE BUTTON, COMPOUND_TEXT e.t.c.  The image below will help with the three attribute addition requirements. 
weight to row

**Acceptance Criteria**
- [ ] The service button should have the described UI/UX
- [ ] It should be possible to set the enable attribute on the fab button and the extended actions. 
- [ ] It should be possible to add icons to the fab button and extended actions.  
- [ ] The ROW view should have a weight attribute. 
- [ ] We should have a vertical spacer view.
- [ ] We should be able to set the border radius to the relevant fhircore configuration views. 

Continuation: https://github.com/opensrp/fhircore/issues/1788

### LINKED ISSUE #1788: [Phase 2] UI/UX and functionality updates for the configuration views.
- [x] Add a register cache refresh action on the questionnaire onClose action. 
- [x] Add the ability to show SnackBars to FHIR Core. These are potential replacements for the Toasters
    - [x] Make the SnackBar dismiss time configurable. 
snack-bar-sample

- [x]  The `menuActionButton` on the Navigation should be hidden when the `visible` attribute is set to `false`.

Screenshot from 2022-10-29 16-03-09

- [x] Add the ability to configure different report configurations for different configuration menus. The screenshot below shows a situation where we can have 2 reports menu items in the same navigation drawer. 

Screenshot from 2022-10-29 16-07-57

- [x] Add an action attributes to the `COMPOUND_TEXT` view. 
    - [x] This should be able to open a register and profile.
    - [ ] Count the number of results in the register the action is to open. This should similar to the count used on the Navigation configuration. 

Screenshot from 2022-10-29 16-21-30

- [x] Add the ability to define the Register title (used on the top) on the register configs. Right the registers use the navigation item `display` item for display. We can make the `display` attribute on the register the default title. This way the `COMPOUND_TEXT` on click loading a register doesn't have to care about setting. 
     - [ ] Add the way to pass extra data to add the Register title. e.g when a CHW is viewing a household created by a specific CHA we would want to show the CHA's name on the Register title. 

Screenshot from 2022-10-29 16-22-50

- [x] Add an `enable` attribute to all the views we have on the FHIR core app. When a view is set as disabled the following should happen. 
    - [x] By default the view is not clickable. However, we can configure the onClick behavior. Some disabled views might be clickable. e.g On the Family profile a deceased family member is shown as disabled but we should be able to click on the row and launch their profile.

ellykits:

Fixes #1761 
Fixes #1788 

**Checklist**


ellykits:
Refactored the `ViewRenderer` in this PR to fix the following and add these extra features:
1. Fixed rendering views on ROWs and COLUMNs (Rows and Columns were not rendering) 
2. Include `weight` attribute to assist with positioning of views in parent layout. The view size will be generated as a percentage of the the available screen size. 
3. Split `ViewGroupProperties` into `ColumnProperties` and `RowProperties` the both have different attributes 
4. Fixed ActionableButton  background color
5. Fixed ServiceButton border color 
6. Fixed icons not showing on Fabs and Drawer menu Item (Other clients option)
7. Add border radius to all Properties as well as other common attributes
8. Introduce a disable property for views to mark them as disabled

android/quest/build.gradle
```diff
@@ -138,7 +138,8 @@ android {
             applicationIdSuffix ".ecbis"
             versionNameSuffix "-ecbis"
             versionCode 4
-            versionName "0.1.5-rc1"
+            versionName "0.1.6-rc1"
+            buildConfigField "String[]", "SYNC_STRATEGIES", '{"Location", "Organization"}'
```
FikriMilano, line 142: Can we move this to the ApplicationConfig?
For reference https://github.com/opensrp/fhircore/issues/1778

android/quest/src/main/assets/configs/app/sync_config.json
```diff
@@ -17,13 +17,11 @@
...
-        "name": "organization",
-        "code": "managing-entity.patient.organization",
+        "name": "_id",
+        "code": "_id",
         "base": [
           "Group"
-        ],
-        "type": "token",
-        "expression": "#organization"
+        ]
```
FikriMilano, line 24: Is this intentional? and what's the reason for this change?

FikriMilano (changes requested):

android/quest/src/main/assets/configs/app/sync_config.json
ellykits: Yea this was done intentionally (as a workaround) so as to sync all the Group resources including those used in stock.

pld (approved):

pld:
@FikriMilano can you check to see if the changes you requested have been addressed? thank you!

dubdabasoduba:
@ellykits What is remaining here? I would like to merge this to `main` then we can create issues for the tests and minor bugs. 

@pld @f-odhiambo

PR #1791, opened 2022-10-31
1756 create month year selection

### LINKED ISSUE #1756: [In App Reporting] Create a month selection interface for period selection (Start/End date)

ahsanbhatti49:

Fixes #1756

**Checklist**


android/quest/src/main/res/drawable/ic_grey_down_arrow.xml
```diff
@@ -0,0 +1,10 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="14dp"
+    android:height="8dp"
+    android:viewportWidth="14"
+    android:viewportHeight="8">
+  <path
+      android:pathData="M0.293,0.293C0.683,-0.098 1.317,-0.098 1.707,0.293L7,5.586L12.293,0.293C12.683,-0.098 13.317,-0.098 13.707,0.293C14.098,0.683 14.098,1.317 13.707,1.707L7.707,7.707C7.317,8.098 6.683,8.098 6.293,7.707L0.293,1.707C-0.098,1.317 -0.098,0.683 0.293,0.293Z"
+      android:fillColor="#919191"
+      android:fillType="evenOdd"/>
+</vector>
```
ellykits, line 10: I think we can use compose icon here instead.

android/quest/src/main/res/drawable/ic_grey_right_arrow.xml
```diff
@@ -0,0 +1,10 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="8dp"
+    android:height="14dp"
+    android:viewportWidth="8"
+    android:viewportHeight="14">
+  <path
+      android:pathData="M0.293,0.293C0.683,-0.098 1.317,-0.098 1.707,0.293L7.707,6.293C8.098,6.683 8.098,7.317 7.707,7.707L1.707,13.707C1.317,14.098 0.683,14.098 0.293,13.707C-0.098,13.317 -0.098,12.683 0.293,12.293L5.586,7L0.293,1.707C-0.098,1.317 -0.098,0.683 0.293,0.293Z"
+      android:fillColor="#919191"
+      android:fillType="evenOdd"/>
+</vector>
```
ellykits, line 10: I think we can use relevant compose `Icon` instead. You can define the tint color as desired.

android/quest/src/main/res/values/colors.xml
```diff
@@ -10,6 +10,9 @@
     <color name="dodger_blue">#1976D2</color>
     <color name="gray">#FF808080</color>
     <color name="light_gray">#FFD3D3D3</color>
+    <color name="other_light_gray">#F7F7F7</color>
+    <color name="dark_gray">#6F7274</color>
```
ellykits, line 14: You can instead use `DefaultColor` from `Colors.kt` and change the transparency e.g.

```kotlin 
DefaultColor.copy(alpha = 0.5) // Moderately transparent
DefaultColor.copy(alpha = 0.2) // Highly transparent
```

No need for defining same color code separately with different shades.

android/quest/src/test/assets/configs/app/measure_report_config.json
```diff
@@ -2,6 +2,7 @@
   "appId": "app",
   "configType": "measureReport",
   "registerId": "ancRegister",
+  "registerDate": "2020-10-27",
```
ellykits, line 5: @f-odhiambo Are we suppose to refer to this as `registerDate`?

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/screens/ReportTypeSelectorScreen.kt
```diff
@@ -77,26 +74,12 @@ fun ReportTypeSelectorScreen(
...
     screenTitle = screenTitle,
-    dateRange = measureReportViewModel.dateRange,
-    onDateRangeSelected = { newDateRange ->
-      measureReportViewModel.onEvent(MeasureReportEvent.OnDateRangeSelected(newDateRange))
+    onGenerateReportClicked = { date ->
+      measureReportViewModel.onEvent(
+        MeasureReportEvent.GenerateReport(navController, context),
+        date
+      )
     },
```
ellykits, line 83: I do not think we needed to refactor the existing measure reporting implementation. In app reporting should be implemented and configured independently. Should have it's own UI, data models, repository and viewmodel.

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/components/DateSelectionBox.kt
```diff
@@ -133,6 +136,31 @@ fun showDateRangePicker(
...
+          if ("mDaySpinner" == datePickerField.name) {
+            datePickerField.isAccessible = true
+            val dayPicker = datePickerField[datePicker]
+            (dayPicker as View).visibility = View.GONE
+          }
+        }
+      }
+    }
+  } catch (ex: Exception) {
+    ex.printStackTrace()
```
ellykits, line 159: Use `Timber` instead to log errors. It would be great not to generalize the exception caught here. You can catch the specific exception instead.

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/components/DateSelectionBox.kt
```diff
@@ -133,6 +136,31 @@ fun showDateRangePicker(
...
+        val datePickerFields = datePickerDialogField.type.declaredFields
+        for (datePickerField in datePickerFields) {
+          if ("mDaySpinner" == datePickerField.name) {
+            datePickerField.isAccessible = true
+            val dayPicker = datePickerField[datePicker]
+            (dayPicker as View).visibility = View.GONE
+          }
+        }
+      }
+    }
```
ellykits, line 157: Any reason for using reflection in this implementation?

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/MeasureReportViewModel.kt
```diff
@@ -245,10 +264,7 @@ constructor(
               launchSingleTop = true
             }
           }
-          .onFailure {
-            Timber.w(it)
-            toggleProgressIndicatorVisibility(false)
-          }
+          .onFailure { toggleProgressIndicatorVisibility(false) }
```
ellykits, line 267: The error log is omitted with this new update.

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/screens/ReportTypeSelectorScreen.kt
```diff
@@ -154,191 +136,99 @@ fun ReportTypeSelectorPage(
...
-                verticalArrangement = Arrangement.Bottom
-              ) {
-                Row(modifier = modifier.fillMaxWidth(), verticalAlignment = Alignment.Bottom) {
-                  GenerateReportButton(
-                    generateReportEnabled = generateReport,
-                    onGenerateReportClicked = onGenerateReportClicked
-                  )
-                }
-              }
-            }
```
ellykits, line 192: This implementation is required for measure reporting but has been deleted.

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/MeasureReportViewModel.kt
```diff
@@ -266,9 +282,7 @@ constructor(
           end = endDateFormatted,
           reportType = POPULATION,
           subject = null,
-          practitioner = null
-          /* TODO DO NOT pass this id to MeasureProcessor as this is treated as subject if subject is null.
-          practitionerId?.asReference(ResourceType.Practitioner)?.reference*/ ,
+          practitioner = practitionerId?.asReference(ResourceType.Practitioner)?.reference,
```
ellykits, line 285: There is a `TODO` comment above this line instructing not to provide a non-null value as `practitionerId`. There is a possibility with this new change to have a non-nullable `practitionerId`. Was the bug resolved in `MeasureProcessor`?

android/quest/src/main/java/org/smartregister/fhircore/quest/data/report/measure/MeasureReportRepository.kt
```diff
@@ -61,4 +61,9 @@ constructor(
       registerId = measureReportConfiguration.registerId
     )
   }
+
+  /** returns start date of campaign to get the month/year list start month */
+  fun getCampaignStartDate(): String {
+    return measureReportConfiguration.registerDate
+  }
```
ellykits, line 68: View model should have access to the configuration model. I think it is not necessary to have this in the repository class.

ellykits (changes requested):

ellykits:
@f-odhiambo @dubdabasoduba The implementation in this PR is refactoring the current Measure reporting code. I think in-app reporting should be independent of the current measure reporting implementation. So the existing code should not be refactored. We should however implement new UI/ViewModels, models/Repository for the in-app reporting.

dubdabasoduba:
> @f-odhiambo @dubdabasoduba The implementation in this PR is refactoring the current Measure reporting code. I think in-app reporting should be independent of the current measure reporting implementation. So the existing code should not be refactored. We should however implement new UI/ViewModels, models/Repository for the in-app reporting.

@ellykits @f-odhiambo @ahsanbhatti49  This is my thinking on in-app reports. 

**UI Pieces** 
---

- These should be independent of measure reports. i.e each widget should define the input it needs for it to display data. 
- This should allow for the following
    - A way to define the modules for the indicators. This is basically a grouping of indicators. This should be configurable.  
        - Make sure the page can be hidden. 
        - The list of modules is defined on a JSON that is uploaded as a Binary on the FHIR server.  
    - Date selection. We have 2 options here. 
        - Predefined date selection. This is where we list the years in expandable views that hold a predefined period. The period can be months or quarters. 
            - This should also be configurable using a JSON uploaded as a Binary on the FHIR server. 
        - Non-predefined date selection. Here we show the user the date picker widget and allow them to pick the periods.  
    - A page to show all the views displaying the data from the different indicator aggregations. This should also be configurable.  
        - Have the ability to define 1 or multiple views on one page. For each view definition, the following should be possible. 
            - Definition of the VIEW (widget) to display the data.  
            - Definition of the relevant attributes of the VIEW (widget) e.g titles, post & prefixes 
            - Definition of where to find the aggregated data. This is the piece that defines how the VIEW (widget) interacts with the Middleware below. 
            - All the text here should be **translatable**. 

**Middleware (Data transformer)**
---

- This is the layer that should interact with the different technologies we are using to aggregate data on the app. We have two options for data aggregation technologies. 
    - Measures -> MeasureReports (This is the recommended FHIR approach). This is what FHIR core is defaulting to use. 
    - SQL queries that evaluate and aggregate the data in the JSONs. 
        - FHIR core will only use this if the above FHIR recommended method doesn't work.  
- This piece is responsible for providing the VIEW (widget) with the correctly formatted data it needs for display.  


**Aggregation technology**
--- 

- Here we will hold the code responsible for the data aggregation for the specific technology we use. i.e this is where we have all the code that defines how the Measure or Custom queries are executed to generate MeasureReports and a list of indicator data respectively. 


Some of the UI pieces will determine when and for which indicators we can run period reports generation e.g if the users have a predefined date selection then it's easy to anticipate the data they need to see and then trigger a background task to generate reports.  

If the users don't have predefined date selections then we can only trigger the background report generation after the date selection. This means we have to block the users from using the reports page until the reports are generated. 

I realize I might be mumbling so feel free to set up a call in case you guys need more clarifications.  

cc: @pld

ellykits:
Labeling this with DNM tag until the discussion above is resolved.

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/screens/ReportTypeSelectorScreen.kt
ahsanbhatti49: I have refactored the code as we have to reuse some code from existing viewmodel. 
 @f-odhiambo  Please let me know if need to keep previous ui as it is or not ?

android/quest/src/main/res/drawable/ic_grey_down_arrow.xml
ahsanbhatti49: Removed this resource

android/quest/src/main/res/drawable/ic_grey_right_arrow.xml
ahsanbhatti49: Removed this resource

android/quest/src/main/res/values/colors.xml
ahsanbhatti49: thanks @ellykits  for this practice.

android/quest/src/test/assets/configs/app/measure_report_config.json
ahsanbhatti49: This is the date which will be used as campaign register date,

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/components/DateSelectionBox.kt
ahsanbhatti49: done with this .

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/components/DateSelectionBox.kt
ahsanbhatti49: Sorry my bad , I used this code for sample used in local branch, And Mistakenly merged in this branch.
I have removed this.

pld:
> > @f-odhiambo @dubdabasoduba The implementation in this PR is refactoring the current Measure reporting code. I think in-app reporting should be independent of the current measure reporting implementation. So the existing code should not be refactored. We should however implement new UI/ViewModels, models/Repository for the in-app reporting.
> 
> @ellykits @f-odhiambo @ahsanbhatti49 This is my thinking on in-app reports.

**UI Pieces** 
---

- These should be independent of measure reports. i.e each widget should define the input it needs for it to display data. 
- This should allow for the following
    - A way to define the modules for the indicators. This is basically a grouping of indicators. This should be configurable.  
        - Make sure the page can be hidden. 
        - The list of modules is defined on a JSON that is uploaded as a Binary on the FHIR server.  
    - Date selection. We have 2 options here. 
        - Predefined date selection. This is where we list the years in expandable views that hold a predefined period. The period can be months or quarters. 
            - This should also be configurable using a JSON uploaded as a Binary on the FHIR server. 
        - Non-predefined date selection. Here we show the user the date picker widget and allow them to pick the periods.  
    - A page to show all the views displaying the data from the different indicator aggregations. This should also be configurable.  
        - <s>Have the ability to define 1 or multiple views on one page. For each view definition, the following should be possible. 
            - Definition of the VIEW (widget) to display the data.  
            - Definition of the relevant attributes of the VIEW (widget) e.g titles, post & prefixes 
            - Definition of where to find the aggregated data. This is the piece that defines how the VIEW (widget) interacts with the Middleware below. 
            - All the text here should be **translatable**.</s>

**Middleware (Data transformer)**
---

- This is the layer that should interact with the different technologies we are using to aggregate data on the app. We have two options for data aggregation technologies. 
    - Measures -> MeasureReports (This is the recommended FHIR approach). This is what FHIR core is defaulting to use. 
    - <s>SQL queries that evaluate and aggregate the data in the JSONs. 
        - FHIR core will only use this if the above FHIR recommended method doesn't work.  
- This piece is responsible for providing the VIEW (widget) with the correctly formatted data it needs for display.  <


**Aggregation technology**
--- 

- Here we will hold the code responsible for the data aggregation for the specific technology we use. i.e this is where we have all the code that defines how the Measure or Custom queries are executed to generate MeasureReports and a list of indicator data respectively. 


Some of the UI pieces will determine when and for which indicators we can run period reports generation e.g if the users have a predefined date selection then it's easy to anticipate the data they need to see and then trigger a background task to generate reports.  

If the users don't have predefined date selections then we can only trigger the background report generation after the date selection. This means we have to block the users from using the reports page until the reports are generated.</s> 

I realize I might be mumbling so feel free to set up a call in case you guys need more clarifications.  

> cc: @pld 
---

This makes sense, but for this first iteration we want to move the custom queries and aggregation out of scope. We can discuss that more and come back to it. I added strike through above to show what I'm referring to as out of scope.

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/MeasureReportViewModel.kt
```diff
@@ -321,7 +343,8 @@ constructor(
                 )
               }
             }
-        // if each stratum evaluated to single item, display all under one group else for each add a
+        // if each stratum evaluated to single item, display all under one group else for each add
+        // a
```
pld, line 347: I like these comments 👍 you can pull line 348 into the comment on line 347

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/MeasureReportViewModel.kt
```diff
@@ -360,6 +389,27 @@ constructor(
       }
   }
 
+  /** Returns a Map of year-month list for for all months falling in given measure period */
```
pld, line 392: can you make this a proper kdoc string please? https://kotlinlang.org/docs/kotlin-doc.html#kdoc-syntax

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/MeasureReportViewModel.kt
```diff
@@ -360,6 +389,27 @@ constructor(
...
+
+    while (lastDate!!.after(startDate)) {
+      yearMonths.add(ReportRangeSelectionData(lastDate.asMmmm(), lastDate.asYyyy(), lastDate))
+
+      lastDate = lastDate.plusMonths(-1)
+    }
+    return yearMonths.toList().groupBy { it.year }
+  }
+
+  /** * show Fixed month/year listing */
```
pld, line 410: can we make this a full docstring as well, I know we haven't consistent in the past but we want to try and do better going forward :)

pld:
removing the dnm tag, we're aligned on the mvp approach hitting the points from my comments on Ben's comments, @ellykits if there's something we missed please comment and re-add dnm tag

android/engine/src/test/java/org/smartregister/fhircore/engine/util/extension/DateTimeExtensionTest.kt
```diff
@@ -75,6 +75,14 @@ class DateTimeExtensionTest : RobolectricTest() {
 
     assertEquals("Feb", result)
   }
+  @Test
+  fun `Date asMmmm() should return correct formatted date`() {
+    val date = DateUtils.parseDate("2022-03-02", "yyyy-MM-dd")
+
+    val result = date.asMmmm()
+
+    assertEquals("March", result)
```
ellykits, line 84: Have you verified the returned date format is translatable?

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/screens/MeasureReportResultScreen.kt
```diff
@@ -81,6 +82,7 @@ fun MeasureReportResultScreen(
   )
 }
 
+@SuppressLint("UnusedMaterialScaffoldPaddingParameter")
```
ellykits, line 85: Apply the padding as suggested, remove this suppression. See how this is used in [RegisterScreen#L111](https://github.com/opensrp/fhircore/blob/7781e177fc4a60216172d1893fa64c883b1f8b82/android/quest/src/main/java/org/smartregister/fhircore/quest/ui/register/RegisterScreen.kt#L111)

android/quest/src/test/java/org/smartregister/fhircore/quest/app/fakes/Faker.kt
```diff
@@ -96,4 +96,31 @@ object Faker {
...
+    runBlocking {
+      configurationRegistry.loadConfigurations(
+        appId = APP_DEBUG,
+        context = ApplicationProvider.getApplicationContext()
+      ) {}
+    }
+
+    return configurationRegistry
+  }
 }
```
ellykits, line 126: This is a duplicate of [Faker#L33](https://github.com/opensrp/fhircore/blob/7781e177fc4a60216172d1893fa64c883b1f8b82/android/quest/src/androidTest/java/org/smartregister/fhircore/quest/Faker.kt#L33)

ellykits (changes requested):

android/engine/src/test/java/org/smartregister/fhircore/engine/util/extension/DateTimeExtensionTest.kt
ellykits: As in are the alphabetic words translated when you switch language for the application or update the device language?

ellykits: I think they should but kindly confirm.  Otherwise that can be addressed later shouldn't hinder us from merging this PR.

ellykits (approved):

ellykits:
I've updated the base branch to resolve the merge conflicts in `configurable-ui-updates` branch.

Issue #1814, opened 2022-11-04
[App configuration] Add the ability to download a List resource and the referenced resources from the Composition resource. 


dubdabasoduba:
**Name of feature to enhance**
- The Composition resource is used as the Application manifest. This means all the application-related configurations and common resource types are supposed to be referenced here. 

**Description of feature**
- [ ] Add the ability to download a list of resources if referenced on a composition resource, 
- [ ] Add the ability to download all the resources referenced in the List resource download above. 
    - In the example below the Groups referenced should be downloaded.  

**Acceptance criteria*
- [ ] Download a List resource referenced on the Composition resource then download all the resources referenced on the List resource. 

**Sample List resource**
```json
{
    "resourceType": "List",
    "id": "ea15c35a-8e8c-47ce-8122-c347cefa1b4a",
    "identifier": [
        {
            "use": "official",
            "value": "ea15c35a-8e8c-47ce-8122-c347cefa1b4a"
        }
    ],
    "status": "current",
    "mode": "working",
    "title": "Supply Chain commodities",
    "code": {
        "coding": [
            {
                "system": "http://ona.io",
                "code": "supply-chain",
                "display": "Supply Chain Commodity"
            }
        ],
        "text": "Supply Chain Commodity"
    },
    "entry": [
        {
            "item": {
                "reference": "Group/6f3980e0-d1d6-4a7a-a950-939f3ca7b301"
            }
        },
        {
            "item": {
                "reference": "Group/e50eb835-7827-4001-b233-e1dda721d4e8"
            }
        },
        {
            "item": {
                "reference": "Group/90b10fdb-592c-47b6-a265-c8806a15d77c"
            }
        },
        {
            "item": {
                "reference": "Group/dde1cd4f-bef4-4d2b-ad1b-f63b639ed254"
            }
        }
    ]
}
```

- https://fhir.labs.smartregister.org/fhir/List/ea15c35a-8e8c-47ce-8122-c347cefa1b4a
- http://ecbis-preview.smartregister.org/fhir/List/ea15c35a-8e8c-47ce-8122-c347cefa1b4a

dubdabasoduba:
@qaziabubakar-vd please add an LOE to this issue

qaziabubakar-vd:
@dubdabasoduba Looking at the current composition configs, can we include the list resources in the following way:

`{
      "title": "Lists",
      "mode": "working",
      "section": [
        {
          "title": "Supply Chain commodities",
          "focus": {
[...]

PR #1824, opened 2022-11-09
Issue 1790, 1814 Downloading resources from Composition config

### LINKED ISSUE #1790: [App Configuration] Move the download of the application specific resources  to the Composition resource.
- The Composition resource is used as the Application manifest. This means all the application-related configurations are supposed to be referenced here. 
- The download of these resources should only be triggered from this resource. 
- Right now we have some application-specific configs referenced on the `sync_config.json`.  We need to move these configurations and reference them on the Composition resource. 

**Description of feature**
- [x] Make sure the following resources can be downloaded from the composition resource reference, 
    - [x] Binary Resource. 
    - [x] Questionnaire Resource. 
    - [x] PlanDefinition Resource. 
    - [x] StructureMap Resource. 
    - [x] Library Resource. 
    - [x] Measure Resource. 

**Describe the enhancement**
- [x]  Ability to specify the resources mentioned above on the Composition Resource and download them for usage on the app. 
   - [x] This means removing the references on the `sync_config.json` and sill being able to download the resources since they are referenced on the  Composition resource.

qaziabubakar-vd:

Fixes #1790 #1814

**Checklist**


ellykits (dismissed):

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/main/AppMainViewModel.kt
```diff
@@ -115,6 +127,56 @@ constructor(
...
+          val resourceUrlPath =
+            entry.key +
+              "?${Composition.SP_RES_ID}=$ids" +
+              "&_count=${configService.provideConfigurationSyncPageSize()}"
+
+          Timber.d("Fetching config details $resourceUrlPath")
+
+          oAuthService.getResource(resourceUrlPath).entry.forEach { bundleEntryComponent ->
+            when (bundleEntryComponent.resource) {
+              is ListResource -> {
```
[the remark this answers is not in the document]
ellykits: Update the existing implementation accordingly.

android/engine/src/main/java/org/smartregister/fhircore/engine/data/remote/auth/FhirOAuthService.kt
```diff
@@ -0,0 +1,26 @@
...
+package org.smartregister.fhircore.engine.data.remote.auth
+
+import org.hl7.fhir.r4.model.Bundle
+import retrofit2.http.GET
+import retrofit2.http.Url
+
+interface FhirOAuthService {
+
+  @GET suspend fun getResource(@Url url: String): Bundle
+}
```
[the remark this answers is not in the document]
ellykits: Reuse the existing implementation

android/engine/src/main/java/org/smartregister/fhircore/engine/di/NetworkModule.kt
```diff
@@ -91,9 +91,9 @@ class NetworkModule {
   @Provides
   fun provideFhirResourceService(
     parser: IParser,
-    @OkHttpClientQualifier okHttpClient: OkHttpClient,
+    @AuthOkHttpClientQualifier okHttpClient: OkHttpClient,
```
ellykits, line 94: There was a reason for providing two different `OkHttpClient` instances, this was to limit the logging for different requests another was to use different interceptors to set the Bearer access token for particular requests. If those are no longer the intentions then perhaps we just need a single instance of the `OkHttpClient` thus the Qualifiers aren't necessary.

qaziabubakar-vd: Sure, we can use one instance for `OkHttpClient` as we need to have bearer access token for some calls in the `FhirResourceService` for fetching resources like `StructureMaps` and `Questionnaires`

android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/ConfigurationRegistry.kt
```diff
@@ -320,15 +321,46 @@ constructor(
...
+                fhirResourceDataSource.loadData(searchPath).entry.forEach { bundleEntryComponent ->
+                  when (bundleEntryComponent.resource) {
+                    is ListResource -> {
+                      addOrUpdate(bundleEntryComponent.resource)
+                      val list = bundleEntryComponent.resource as ListResource
+                      list.entry.forEach { listEntryComponent ->
+                        /*Here we extract the keys and ids for the resources listed in
+                        the List resource */
+                        val resourceKey = listEntryComponent.item.reference.substringBeforeLast("/")
+                        val resourceId = listEntryComponent.item.reference.substringAfterLast("/")
```
dubdabasoduba, line 347: @ellykits @qaziabubakar can we have a utils function to do this? I assume this is something we will wan to do a lot

dubdabasoduba (dismissed):

pld (dismissed):

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/main/AppMainActivity.kt
```diff
@@ -104,6 +106,13 @@ open class AppMainActivity : BaseMultiLanguageActivity(), QuestionnaireHandler,
       }
     }
 
+    // Register sync listener then run sync in that order
+    syncListenerManager.registerSyncListener(this, lifecycle)
+
+    CoroutineScope(dispatcherProvider.io()).launch {
```
ellykits, line 112: This CoroutineScope is not necessary. The function `fetchNonWorkflowConfigResources` already has a `CoroutineScope` running the code in IO thread. Switching to another context here is not great for performance.

ellykits: Can we call this functional immediately after login before even launching the AppMainActivity? Somewhere in the `QuestLoginService` class?

ellykits: Reason for above the main activity will load the register, and the register has a button for registering client that launches a questionnaire that should have been loaded even when there are no register entries. It would actually be great to display the button only until the download is complete.

ellykits (changes requested):

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/main/AppMainActivity.kt
ellykits: CC @dubdabasoduba

dubdabasoduba: @ellykits This should be loaded immediately after login. Before we start syncing anything else. I am okay with syncing it on the `QuestLoginService`.

ellykits (approved):

PR #1826, opened 2022-11-10
Add FHIRTaskExpireUtil to fetch and mark expired tasks

### LINKED ISSUE #1703: [Quest] Write a service to mark Tasks as expired.
**Enhancement context**
- We currently, only show the `overdue` tasks. However, some tasks get expired after a certain period of them being `overdue`
- These expired tasks can't be performed anymore. We will be tracking the expired tasks by updating the `Task.status` to `cancelled`. 
- We will be tracking the expiry date from the `Task.restriction.period.end`. If the end date is =< `today()` then the task is expired. 

**Implementation**
- [x] Write a service to mark tasks as expired. The timing of the service should be at least after every 1 hour. 
   - [x] The timing should be configurable on the `application_config`.  In case it's not added then the default is set to 1 hour. 
- [x] The `Task.restriction.period.end` should be used to determine the task expiry. 
   - [x] If the `Task.restriction.period.end` date =< `today()` then update the `Task.status` to `cancelled` 

**Acceptance Criteria**
- [x] All the expired tasks should have a status `cancelled`. 
- [x] The service to update the statuses should run periodically.

ekigamba:

Fixes #1703

**Checklist**


android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirTaskExpireUtil.kt
```diff
@@ -0,0 +1,76 @@
...
+import org.smartregister.fhircore.engine.util.extension.isPastExpiry
+import org.smartregister.fhircore.engine.util.extension.toCoding
+import timber.log.Timber
+
+class FhirTaskExpireUtil(val appContext: Context) {
+
+
+  internal fun getFhirEngine() : FhirEngine {
+    return FhirEngineProvider.getInstance(appContext)
+  }
```
ellykits, line 33: I think we can inject the `FhirEngine` dependency instead.

android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirTaskExpireUtil.kt
```diff
@@ -0,0 +1,76 @@
...
+
+
+  suspend fun fetchOverdueTasks() : List<Task> {
+    Timber.i("Start fetching overdue tasks")
+
+    val fhirEngine = getFhirEngine()
+
+    // TODO: Limit & page the results for better performance
```
ellykits, line 41: Will using the `paging` library solve the issue with pagination?

ellykits:
I think we need to move this implementation here [FhirTaskPlanWorker](https://github.com/opensrp/fhircore/blob/d2ab5130917cf1df2aa1ea189133bff380827663/android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirTaskPlanWorker.kt#L35)

This is a job that can be scheduled to update the statuses of all tasks.

CC: @dubdabasoduba @maimoonak

ekigamba:
@ellykits I see that the implementation provided runs the job every 12 hours while the requirements for this job talked about every 1 hour [here](https://github.com/opensrp/fhircore/issues/1703#issue-1389996502). Otherwise, the `FhirTaskExpireUtil` is not a job and only contains the functionality so we can call it from the `FhirTaskPlanWorker` job

dubdabasoduba:
> @ellykits I see that the implementation provided runs the job every 12 hours while the requirements for this job talked about every 1 hour [here](https://github.com/opensrp/fhircore/issues/1703#issue-1389996502). Otherwise, the `FhirTaskExpireUtil` is not a job and only contains the functionality so we can call it from the `FhirTaskPlanWorker` job

@ellykits @ekigamba Can we write the Worker to allow a custom run period set up? i.e we make it possible for each job to define its own execution period.

ellykits:
Sure I meantioned this in another PR by @ahsanbhatti49

f-odhiambo:
@dubdabasoduba Ahsan is working on it on background Tasks so these should be added as Jobs to it

ahsanbhatti49:
> > @ellykits I see that the implementation provided runs the job every 12 hours while the requirements for this job talked about every 1 hour [here](https://github.com/opensrp/fhircore/issues/1703#issue-1389996502). Otherwise, the `FhirTaskExpireUtil` is not a job and only contains the functionality so we can call it from the `FhirTaskPlanWorker` job
> 
> @ellykits @ekigamba Can we write the Worker to allow a custom run period set up? i.e we make it possible for each job to define its own execution period.

We can make a periodic worker that can run after a specific time interval. and define this interval in the config. 
Interval can be specified by Sec,Minutes,hours etc. Lets say if we want to run a job after every 24 hours we can do that. But there is no direct way of specifying that a job should run daily at specific time. for example: daily at 5pm. etc.

android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirTaskExpireJob.kt
```diff
@@ -0,0 +1,92 @@
...
+import dagger.hilt.android.EntryPointAccessors
+import java.util.concurrent.TimeUnit
+import org.smartregister.fhircore.engine.di.FhirTaskExpireEntryPoint
+import org.smartregister.fhircore.engine.util.SharedPreferencesHelper
+
+class FhirTaskExpireJob constructor(val context: Context, workerParams: WorkerParameters) :
+  CoroutineWorker(context, workerParams) {
+
+  var fhirEngine: FhirEngine = getFhirEngine(context)
+  var fhirTaskExpireUtil: FhirTaskExpireUtil = getFhirTaskExpireUtil(context)
```
ellykits, line 35: It's possible to inject dependencies to `CoroutineWorker`. See [AppSyncWorker](https://github.com/opensrp/fhircore/blob/main/android/engine/src/main/java/org/smartregister/fhircore/engine/sync/AppSyncWorker.kt) for example.

android/engine/src/main/java/org/smartregister/fhircore/engine/di/FhirTaskExpireEntryPoint.kt
```diff
@@ -0,0 +1,33 @@
...
+
+/** Created by Ephraim Kigamba - nek.eam@gmail.com on 05-12-2022. */
+@InstallIn(SingletonComponent::class)
+@EntryPoint
+interface FhirTaskExpireEntryPoint {
+
+  fun fhirEngine(): FhirEngine
+
+  fun fhirTaskExpireUtil(): FhirTaskExpireUtil
+}
```
ellykits, line 33: Was this required for dependency injection? Dependencies can be injected to a subclass  of `CoroutineWorker` as constructor parameters.

android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirTaskExpireJob.kt
```diff
@@ -0,0 +1,92 @@
...
+      if (currVersion != version) {
+        existingWorkPolicy = ExistingPeriodicWorkPolicy.REPLACE
+        sharedPreferencesHelper.write(FHIR_TASK_EXPIRE_JOB_VERSION, version)
+      }
+
+      val periodicWorkRequest =
+        PeriodicWorkRequestBuilder<FhirTaskExpireJob>(
+            durationInMins,
+            TimeUnit.MINUTES,
+            5,
```
ellykits, line 74: Do we need a TODO to configure this in future?

ellykits (changes requested):

android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirTaskExpireJob.kt
ekigamba: I believe so but I spent a lot of time trying it but couldn't get it to work. 

You can possibly help to get this working. The current injection works almost similar and depends on Hilt

android/engine/src/main/java/org/smartregister/fhircore/engine/di/FhirTaskExpireEntryPoint.kt
ekigamba: I believe so but I spent a lot of time trying it but couldn't get it to work. 

You can possibly help to get this working. The current injection works almost similar and depends on Hilt

android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirTaskExpireJob.kt
ekigamba: Nice catch. I have updated the code to make the `durationInMins` configurable. The magic number `5` is for the `flexTimeInterval` which I don't expect to change and might be difficult to explain it's impact as a configurable number

android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/app/ApplicationConfiguration.kt
```diff
@@ -36,5 +36,5 @@ data class ApplicationConfiguration(
   val deviceToDeviceSync: DeviceToDeviceSyncConfig? = null,
   val snackBarTheme: SnackBarThemeConfig = SnackBarThemeConfig(),
   val reportRepeatTime: String = "",
-  val taskExpireJobRepeatIntervalConfig: Long = 60
+  val taskExpireJobRepeatInterval: Long = 60
```
pld, line 39: is this in minutes? I think we should add that to the name so it's clear
```suggestion
  val taskExpireJobRepeatIntervalMinutes: Long = 60
```

android/quest/src/main/assets/configs/app/application_config.json
```diff
@@ -6,6 +6,7 @@
   "registerDate": "2020-10-27",
   "reportRepeatTime": "23:22",
   "remoteSyncPageSize": 100,
+  "taskExpireJobRepeatInterval": "15",
```
pld, line 9:
```suggestion
  "taskExpireJobRepeatIntervalMinutes": "15",
```

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/main/AppMainViewModel.kt
```diff
@@ -322,10 +322,10 @@ constructor(
       applicationConfiguration.registerDate
     )
 
-    FhirTaskExpireJob.schedule(
+    FhirTaskExpireWorker.schedule(
       workManager,
       sharedPreferencesHelper,
-      applicationConfiguration.taskExpireJobRepeatIntervalConfig
+      applicationConfiguration.taskExpireJobRepeatInterval
```
pld, line 328:
```suggestion
      applicationConfiguration.taskExpireJobRepeatIntervalMinutes
```

pld (changes requested):

ellykits (dismissed):

pld (approved):

PR #1831, opened 2022-11-11
Questionnaire bug fixes | extraction only on completed

### LINKED ISSUE #1700: [Quest] Update Extraction triggering to only happen for QuestionnaireResponses with the `status` == `Completed`
**Enhancement context**
- We currently, don't check the status of a QuestionnaireResponse before we trigger extraction of any kind. 
- We should only trigger extraction of schedules or to different resources when the Questionnaire is completely filled and the QuestionnaireResponse status is marked as completed. 

**Implementation**
- [x] Add the complete status on the QuestionnaireResponse when the `SAVE` is clicked on the questionnaire. This status should only be added when the QR is not marked as draft. 
- [x] Attach the QR subject before the extraction bit. This should be added immediately after the QR is created and not during the extraction period.  
- [x] Check the QuestionnaireResponse status before passing it through the `extraction` method. 
    - [x] The QuestionnaireResponse status should be set to `Completed` for the extraction to be triggered.  
- [x] If the QuestionnaireResponse status is not `completed` we save the resource as is to the resources table on SQLite. 

**Acceptance Criteria**
- [ ] Extraction of any kind should only happen when the QuestionnaireResponse status is set to `Completed`

maimoonak:
Fixes #1700 

**Checklist**

### DIFF
--- a/android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireActivity.kt
+++ b/android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireActivity.kt
@@ -33,7 +33,10 @@ import com.google.android.fhir.datacapture.QuestionnaireFragment
 import com.google.android.fhir.datacapture.validation.NotValidated
 import com.google.android.fhir.datacapture.validation.QuestionnaireResponseValidator
 import com.google.android.fhir.datacapture.validation.Valid
+import com.google.android.fhir.logicalId
 import dagger.hilt.android.AndroidEntryPoint
+import java.util.Date
+import java.util.UUID
 import javax.inject.Inject
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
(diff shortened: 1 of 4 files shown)


ellykits (commented):
Looks good. Can you verify all the checklisted acceptance criteria on the issue are all met.

maimoonak:
done!

ellykits (approved):

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireActivity.kt
```diff
@@ -247,7 +250,28 @@ open class QuestionnaireActivity : BaseMultiLanguageActivity(), View.OnClickList
   fun getQuestionnaireResponse(): QuestionnaireResponse {
     val questionnaireFragment =
       supportFragmentManager.findFragmentByTag(QUESTIONNAIRE_FRAGMENT_TAG) as QuestionnaireFragment
-    return questionnaireFragment.getQuestionnaireResponse()
+    return questionnaireFragment.getQuestionnaireResponse().apply {
+      // TODO this is required until require condition in [QuestionnaireResponseValidator] is fixed
```
dubdabasoduba, line 254: @maimoonak for the todo here. Do you have an idea of what needs to be done?  If do we have an issue for it? 

@f-odhiambo Can we add this to the SDK sprint?

dubdabasoduba (approved):

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireActivity.kt
maimoonak: @dubdabasoduba @f-odhiambo https://github.com/google/android-fhir/issues/1713

PR #1860, opened 2022-11-24
Mark tasks ready

### LINKED ISSUE #1701: [Quest] Add the ability to mark Tasks are Ready.
**Enhancement context**
- We would like to mark tasks that are ready to be actioned on. 
- These are any tasks that  the `executionPeriod.start` date is <= to today and the `executionPeriod.end` date >= today

**Implementation**
- [x] Write a service that runs after a certain period of time to change the tasks statuses for the Tasks with the status `Requested` to `Ready`  
   - [x]  Tasks only Qualify to move to the `Ready` status if they have the following conditions.
       - [x] The `executionPeriod.start` date is <= to today and the `executionPeriod.end` date >= today.
       - [x] The current status is `Requested` 
- [x] The periodic timing for the service to run should be configurable and set on the `application_config`. The least amount of time should be capped at 1 hour. 
   - [x] In case an implementation does not set this timing on the `application_config` then the app should automatically use the default timing. 

**Acceptance Criteria**
- [ ] Tasks with the status `Requested` should change to `Ready` in least an hour after their creation if the `executionPeriod.start` date is >= to today and the `executionPeriod.end` date < today.

allan-on:

Fixes #1701 

**Checklist**


android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/app/ApplicationConfiguration.kt
```diff
@@ -33,5 +33,6 @@ data class ApplicationConfiguration(
   val syncStrategies: List<String> = listOf(),
   val loginConfig: LoginConfig = LoginConfig(),
   val deviceToDeviceSync: DeviceToDeviceSyncConfig? = null,
-  val snackBarTheme: SnackBarThemeConfig = SnackBarThemeConfig()
+  val snackBarTheme: SnackBarThemeConfig = SnackBarThemeConfig(),
+  val periodicTaskUpdateInterval: Long = 1
```
ellykits, line 37: Consult with @ahsanbhatti49 on the changes requested on this config property. The ideas is to use[ ISO period format ](https://en.wikipedia.org/wiki/ISO_8601#:~:text=ISO%208601%20uses%20the%2024,hour%20between%2000%20and%2024.) to have the flexibility of defining duration

ellykits (dismissed):

android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/app/ApplicationConfiguration.kt
ellykits: We can parse the string using [java.time.Duration#parse](https://developer.android.com/reference/java/time/Duration#parse(java.lang.CharSequence))

android/engine/src/main/java/org/smartregister/fhircore/engine/util/extension/TaskExtension.kt
```diff
@@ -29,4 +29,17 @@ fun Task.hasStarted() =
     this.executionPeriod.hasStart() &&
     with(this.executionPeriod.start) { this.before(today()) || this.isToday() }
 
+fun Task.isReady() =
+  this.hasExecutionPeriod() &&
```
pld, line 33: wouldn't you want the end to be in future? this reads like it's checking that the end is in the past (or is today)

allan-on: Yeah, this makes sense. It wasn't super clear initially from the scoping description in the associated issue.

pld: Yea true good point, I think the issue spec is wrong, the ways tasks will be written it'll be valid between some start and end time (I think? @AnnieMungai @dubdabasoduba you can comment to verify), in the abstract that at least makes sense, I'll correct #1701 too

pld (dismissed):

pld (dismissed):

pld:
@allan-on can you take a look at these failing tests?

android/engine/src/main/java/org/smartregister/fhircore/engine/util/extension/TaskExtension.kt
```diff
@@ -29,6 +29,19 @@ fun Task.hasStarted() =
     this.executionPeriod.hasStart() &&
```
ndegwamartin, line 29: Maybe for this we can re-use the new method `Task.executionStartIsBeforeOrToday`

android/quest/src/test/java/org/smartregister/fhircore/quest/util/mappers/MeasureReportPatientViewDataMapperTest.kt
```diff
@@ -52,7 +52,10 @@ class MeasureReportPatientViewDataMapperTest : RobolectricTest() {
     with(profileViewDataHiv) {
       Assert.assertEquals("TEST_PATIENT", logicalId)
       Assert.assertEquals("Bareera Hadi", name)
-      Assert.assertEquals("23y", age)
+      Assert.assertEquals(
+        "24y",
```
ndegwamartin, line 56: Since we know the actual DOB from the [test data JSON](https://github.com/opensrp/fhircore/blob/main/android/quest/src/test/resources/patient-registration-questionnaire/sample/patient.json) i.e. **1998-12-14** we can make the expected value dynamic by doing something like this

```
LocalDate dob = LocalDate.of(1998, 12, 14);  
Period period = Period.between(dob, LocalDate.now());  
expectedDate = "${period.getYears()}y"
```

allan-on: Yeah,  it makes sense to compute the value of the age from within the test 👍🏾

pld (approved):

PR #1862, opened 2022-11-24
Handle exceptions while performing extraction using StructureMap 

### LINKED ISSUE #1847: [Quest] Allow submitting of a Questionnaire and create a QuestionnaireResponse if StructureMap extension is not provided

**Description of feature**

**Describe the enhancement**

### LINKED ISSUE #1848: [Quest] Provide error-handling if StructureMap has an error during the extraction process
Provide error handling if StructureMap has an error during the extraction process

**Description of feature**
Add a mechanism to prevent a crash when there is an issue with the StructureMap and also log/provide feedback for action

**Describe the enhancement**
- [x] 1. Alert user if an error was encountered 
- [x] 2. Log the details in the app for debugging / troubleshooting purposes

### LINKED ISSUE #1849: [Quest] Provide error-handling if StructureMap extension is provided but the StructureMap resource is not available locally/not yet synced
Provide error handling if StructureMap extension is provided but the StructureMap resource is not available locally/not yet synced

**Description of feature**
The app should not crash upon missing StrcutureMaps for the linked Questionnaire being populated before the extraction process 

**Describe the enhancement**
On submission of a populated Questionnaire
- [x] 1. Generate a QuestionnaireReponse 
- [x] 2. If a StrucutureMap is missing log this in-app for troubleshooting and also  alert the user of the issue

SebaMutuku:

Fixes #1847
Fixes #1848
Fixes #1849 

**Checklist**


dubdabasoduba:
@SebaMutuku Please add the appropriate PR title and link the issue in the description

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireViewModel.kt
```diff
@@ -392,21 +391,41 @@ constructor(
...
+            )
+          }
+        }
+        .onSuccess {
+          bundle = it
+          context.showToast(context.getString(R.string.structure_success))
+        }
+        .onFailure { exception ->
+          Timber.e(exception)
+          if (exception.message!!.contains("StructureMap")) {
```
pld, line 421: what's the exception that gets thrown here? Is there a more precise way to capture this exception? Since it's only checking for a string, couldn't this catch inappropriate exceptions

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireViewModel.kt
```diff
@@ -392,21 +391,41 @@ constructor(
...
+            context.showToast(context.getString(R.string.structure_map_missing_message))
```
pld, line 422: can we mention the specific structuremap it was looking for in the error?

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireViewModel.kt
```diff
@@ -392,21 +391,41 @@ constructor(
...
+          } else {
+            context.showToast(context.getString(R.string.structure_error_message))
```
pld, line 424: can we mention the specific structuremap that had a problem in the error?

android/quest/src/test/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireViewModelTest.kt
```diff
@@ -914,9 +917,52 @@ class QuestionnaireViewModelTest : RobolectricTest() {
...
+      context.getString(R.string.structure_map_missing_message)
+    val context = mockk<Context>(relaxed = true)
+    val questionnaire = Questionnaire()
+    val questionnaireResponse = QuestionnaireResponse()
+
+    coEvery { questionnaireViewModel.retrieveStructureMapProvider() } throws
+      NullPointerException("StructureMap on Null Object")
+
+    questionnaireViewModel.performExtraction(context, questionnaire, questionnaireResponse)
+
```
pld, line 945: let's check that questionnaireResponse is properly filled when missing, that can be in a separate test

android/quest/src/test/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireViewModelTest.kt
```diff
@@ -914,9 +917,52 @@ class QuestionnaireViewModelTest : RobolectricTest() {
...
+    val missingStructureMapExceptionMessage = context.getString(R.string.structure_error_message)
+    val context = mockk<Context>(relaxed = true)
+    val questionnaire = Questionnaire()
+    val questionnaireResponse = QuestionnaireResponse()
+
+    coEvery { questionnaireViewModel.retrieveStructureMapProvider() } throws
+      Exception("Failed to process resources")
+
+    questionnaireViewModel.performExtraction(context, questionnaire, questionnaireResponse)
+
```
pld, line 959: let's check that questionnaireResponse is properly filled when error, that can be in a separate test

pld (changes requested):

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireViewModel.kt
SebaMutuku: `java.lang.NullPointerException:Attempt to invoke StructureMap on null object reference.`

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireViewModel.kt
SebaMutuku: If a `StructureMap` exists but extraction Fails, other exceptions are thrown. This is the reason the second check. To handle that.

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireViewModel.kt
SebaMutuku: Given the above, added a check to verify that exception is an instance of `NullPointerException`

android/quest/src/test/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireViewModelTest.kt
SebaMutuku: The function `performExtraction` doesn't alter the contents of `questionnaireResponse` and we don't need to add it to unit test as part of this PR.

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireViewModel.kt
pld: We want some info so that we can track down what's the causing the problem, maybe the form name would be enough, if we can only have a single structuremap tied to a form, but if we can have more then 1 structuremap tied to a form, we'd want a way to tell which

SebaMutuku: Added questionnaire name to as part of the toast message. This is because in case of exception, the function  `ResourceMapper.extract` returns an empty bundle and hence we cannot know the StructureMap name.

pld (approved):

PR #1886, opened 2022-12-08
Fix issue #1882 , Generate a Unique random number of length 6

### LINKED ISSUE #1882: [Human Readable IDs] Generate random Human readable IDs and pass them to the forms for usage | MVP
**Feature Context**
- OpenSRP maintains a list of unique human-readable IDs assigned to different clients for identification. 
- We want to build an MVP for this on FHIR core. This initial piece will look at generating a set of numbers that can be read by health practitioners to identify the patients. 
- We will then use the generated number on the forms by passing it as follows https://github.com/opensrp/fhircore/blob/main/android/quest/src/main/assets/configs/app/profiles/inventory_profile_config.json#L159-L174

**Feature Implementation**
- [ ] In the [RulesFactory](https://github.com/opensrp/fhircore/blob/main/android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt) Class, write a method that generates a set of whole numbers (min 6).
- [ ] This method should accept a parameter that describes the length of the number.  If the parameter is not available, generate a number with a length of 6. 

**Feature acceptance criteria** 
- [ ]  The method should be able to be called from the configuration rules.  
- [ ] The method should generate numbers with the specified length or a min length of 6 when the length is not user specified.

SebaMutuku:
* Add function to generate a unique number of length 6
* Add tests to confirm the length of the generated number is 6

Fixes #1882 

**Checklist**


ellykits:
Can we use a verifiable number generator based on existing algorithms that can be validated?

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
```diff
@@ -337,12 +337,21 @@ constructor(
     fun formatDate(date: Date, expectedFormat: String = "E, MMM dd yyyy"): String {
       return date.formatDate(expectedFormat)
     }
+
+    fun generateRandomIntLengthSix(): Int {
+      val uniqueId =
+        System.currentTimeMillis().toString().takeLast(6).toInt() +
+          (100000..maxInger).shuffled().random()
+      return if (uniqueId.toString().length > 6) uniqueId.toString().takeLast(6).toInt()
+      else uniqueId
```
pld, line 346: I agree w/@ellykits, we can simplify this. Does the below work? Would be nice to move the constants elsewhere, but it seems odd to have them in the RulesEngine, but guess we could move to the companion object for now

```suggestion
      return Random.ints(inclusiveSixDigitMinimum, exclusiveSixDigitMaximum)
```

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
```diff
@@ -337,12 +337,21 @@ constructor(
     fun formatDate(date: Date, expectedFormat: String = "E, MMM dd yyyy"): String {
       return date.formatDate(expectedFormat)
     }
+
+    fun generateRandomIntLengthSix(): Int {
```
pld, line 341: I think this name is clearer, can you also please add a doc string

```suggestion
    fun generateRandomSixDigitInt(): Int {
```

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
```diff
@@ -337,12 +337,21 @@ constructor(
...
+      else uniqueId
+    }
   }
 
   companion object {
     private const val FHIR_PATH = "fhirPath"
     private const val DATA = "data"
     private const val TRUE = "true"
     private const val SERVICE = "service"
+    private const val maxInger = 999999
```
pld, line 355:
```suggestion
    private const val inclusiveSixDigitMinimum = 100000
    private const val exclusiveSixDigitMaximum = 1000000
```

pld (changes requested):

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
SebaMutuku: @pld this is noted. What are the possibilities of `Random.ints(inclusiveSixDigitMinimum, exclusiveSixDigitMaximum) ` generating same number in two different devices?
That is why I thought of including last digits of timestamp as part of the return value

pld: Yes that's definitely possible, @dubdabasoduba said that this iteration we're ok w/that and we'll have to be sure any users know, so important to note that in the docs

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
ellykits: Conventionally Kotlin constants should be  in [UPPER_SNAKE_CASE](https://developer.android.com/kotlin/style-guide#constant_names)

pld: cool, updated

pld:
For future work, it feels like we have a somewhat arbitrary set of functions in the RuleFactory class, am I missing some pattern or should we reorganize these at some point?

ellykits:
> For future work, it feels like we have a somewhat arbitrary set of functions in the RuleFactory class, am I missing some pattern or should we reorganize these at some point?

We will reorganize the functions in future.

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
ellykits: Great. This is noted @pld

SebaMutuku:
@pld @ellykits I have created an extension file just incase some functions are shared or you want to move the logic there.

pld:
> @pld @ellykits I have created an extension file just incase some functions are shared or you want to move the logic there.

Thanks, let's wait until we thought through the structure before we create this file, I've removed it for now

pld (commented):
I was reading the JavaDocs, Kotlin's random is actually inclusive at the upper bound, sorry for the misdirect, tweaked the code to handle that, we also don't need shuffle if we call random it seems, @SebaMutuku can you add a docstring to `generateRandomSixDigitInt`? Should be good to merge w/that

SebaMutuku:
@pld kdocs added.

pld (dismissed):
We might want to use the Random class in the next version so we can use our own seed and get some predictability in the tests, but for now, lgmt!

pld (approved):

PR #1898, opened 2022-12-13
Update logout flow for when PIN is set/unset

### LINKED ISSUE #1883: [Authentication] Update Logout flow when PIN Login is configured/or not
**Feature Context**
Update the current Logout from the app flow so that the user is able to easily enter back into the application post-logout if PIN authentication is configured or not 

**Feature Implementation**
See the current proposed flow 
Flow chart

NB. Editable [link](https://docs.google.com/drawings/d/1pgHQza5uMhlrnk0GNEqdac7_1t86Gg-gNEvZ449A9r8/edit?usp=sharing) to the above flow chart

**Feature acceptance criteria** 
- [ ] **PIN was configured**: Take the user to enter their 4 Digit PIN
- [ ] **PIN was not configured**  : Take the user to enter their username and password

allan-on:

Fixes #1883  

**Checklist**


android/engine/src/main/java/org/smartregister/fhircore/engine/ui/login/LoginActivity.kt
```diff
@@ -56,7 +56,7 @@ class LoginActivity : BaseMultiLanguageActivity() {
       ) {
         loginService.navigateToPinLogin(false)
       } else if (accountAuthenticator.hasActiveSession() && !isPinEnabled) {
-        loginService.navigateToHome()
+        accountAuthenticator.localLogout()
```
ellykits, line 59: What would happen here is if the application is not using PIN login and there is an active session (active access token) the user will be forced to log out. Is this the desired use case?

ellykits: User will be required to login each time they use the application even if there is already an existing active session. I do not think this is correct.

ellykits (changes requested):

android/engine/src/main/java/org/smartregister/fhircore/engine/ui/login/LoginActivity.kt
allan-on: The changes affect the workflow when the user leaves the app (or the app is closed) and has no effect when the user goes back to the app if it was in the background. 
- Currently, if there's an  active session and the PIN is set, they'll always be taken to the PIN Login screen
- The changes introduce being directed to the Login screen if there's an active session and a PIN isn't enabled/set
This is based on the check `"Was PIN login setup = No"` workflow as detailed in this flow chart  

image

pld: This is a good question, I don't know the answer. @f-odhiambo correct if wrong, but I think the entire diagram above is describing what applies when there is not an active session. I assumed when there is an active session, no login required, but I'm not sure of that.

dubdabasoduba: @pld I think making the user log in every time they open the app they need to authenticate. The reason for this to make sure that the data is not accessed by un-authorised users. 

I think we don't need to care much about the sessions in the authentication flow becuase the auth token renewal process will handle that when the app is trying to sync data to the server.  Might be worng about this.

allan-on:
QA/Acceptance criteria checks based on the implementation:

- [ ] Whether you exit the app (or remove it from recents), OR you log out manually, you'd have to log back in
- [ ] Logging in is either via PIN if configured OR via username + password (regardless of how you exited the app)
- [ ] The PIN Activity/Screen is only shown when the PIN is configured as `enabled` and a PIN is set
- [ ] On login with username and password the access token is only retrieved from the Keycloak endpoint if there's no active session (available access token) otherwise the user is logged in locally
- [ ] The local and Keycloak session is only cleared if you log out while the app/device is online

pld (approved):

PR #1910, opened 2022-12-16
Improve app performance by fixing memory leaks

### LINKED ISSUE #1908: [Performance] Registers take long to refresh after adding new records.
**Describe the bug**
- Sometimes when a new record is added the registers take a while to refresh. e.g if we add a new household to the register will take a while to refresh.  The register refresh also seems to use up a bit of ram which ends up slowing down the device. 
- The same thing happens when you type on the search bar to find something on the register. 

[Screencast from 2022-12-16 11-00-59.webm](https://user-images.githubusercontent.com/4817531/208051947-8f3565f2-c91b-4640-a165-916abf03279b.webm)

[Screencast from 2022-12-16 10-57-33.webm](https://user-images.githubusercontent.com/4817531/208051964-e9f4c672-1a25-4106-9c2d-b42bf8e345b4.webm)

**Bug fix implementation**
- [ ] Update the register and profile refresh to be a bit faster and probably use less ram.  
- [ ] Update the search to ideally search and present the results faster. 
     - [ ] it would be sweet if we attained an O(1) time complexity.

ellykits:

Fixes #1908

**Checklist**


ellykits:
This PR:
- Refactors the content of `ResourceData` (a model that is used to hold register/profile data) to exclude FHIR resources that are not required in Composable views. 
- Fixes issue of memory hoarding each time a register is launched
- Reverts implementation for MeasureReport worker scheduler that was affecting application performance after running the app for a longer period.

ellykits:
See the previously retained size of ResourceData (**approx 31MBs**) in heap memory (while displaying the register screen)

Screenshot from 2022-12-17 10-42-33

This has now been reduced to approx (**0.02MBs**)
Screenshot from 2022-12-17 10-48-34

dubdabasoduba:
@pld I am going to merge this for the rest of the team can build on it. 

@ndegwamartin is investigating why the performance still degrades after an hour of App usage. 
@ahsanbhatti49 is investigating the high resource usage on the scheduled measure reports generation code.

ahsanbhatti49:
@f-odhiambo  @dubdabasoduba  Referring to Schedule measure report generation,
Lets say if we have 10 module and each module has 7 indicators, then for each indicator we need to call evaluateMeasure  for each measure url which can have lets say around 10000 patients data. evaluateMeasure is a long running task and is already running in separate coroutine. i think this will have a minor effect on performance for the first time. Once we have generated report for all the modules/indicator , then the worker will only generate reports for the new indicators or if the the report is not generated for the current month. I think app is not slowing down after some time due to measure report scheduler.

PR #1927, opened 2023-01-02
Stock Consumption and Onhand

maimoonak:
Fixes #https://github.com/opensrp/fhir-resources/issues/549 

**Checklist**


android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/report/measure/MeasureReportConfig.kt
```diff
@@ -25,4 +25,10 @@ data class MeasureReportConfig(
   val description: String = "",
   val url: String = "",
   val module: String = "",
+  val type: MeasureReportConfigType = MeasureReportConfigType.PATIENT,
 )
+
+enum class MeasureReportConfigType {
```
pld, line 31: can you please propose a way to refactor this that remove this type and the conditional associated with it from the hard coding?

maimoonak: Sure, I have some changes to the way indicators are being handled so it would be rendered in a generic way i.e. For Group level resources if type is subject show counts from MeasureReport.contained > Observation The work is under progress and hoping to completely remove the enum by next week.

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/components/MeasureReportPopulationResultView.kt
```diff
@@ -75,7 +76,10 @@ private fun PopulationResultCard(
             textAlign = TextAlign.Start
           )
           Text(
-            text = resultItem.measureReportDenominator.let { if (it == -1) "" else it.toString() },
+            text =
+              resultItem.measureReportDenominator.let {
+                if (it == MEASURE_REPORT_DENOMINATOR_MISSING) "" else it.toString()
```
ahsanbhatti49, line 81: Should not display here empty string. 
Replacing this with "0"

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/models/MeasureReportPopulationResult.kt
```diff
@@ -19,12 +19,14 @@ package org.smartregister.fhircore.quest.ui.report.measure.models
...
+
 @Stable
 @Serializable
 data class MeasureReportPopulationResult(
   val title: String = "",
   val count: String = "",
   val dataList: List<MeasureReportIndividualResult> = emptyList(),
   val indicatorTitle: String = "",
-  val measureReportDenominator: Int = 0
+  val measureReportDenominator: Int = MEASURE_REPORT_DENOMINATOR_MISSING
```
pld, line 31: can we set this to null and have that be the check? I'd worry that someone does math w/this not realizing it's a sentinel

pld (commented):
looks great!

maimoonak:
Done. Kindly review again

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/MeasureReportViewModel.kt
```diff
@@ -360,164 +327,188 @@ constructor(
...
+    subject: String?
+  ): MeasureReport {
+    return fhirOperator.evaluateMeasure(
+      measureUrl = measureUrl,
+      start = startDateFormatted,
+      end = endDateFormatted,
+      reportType = reportType,
+      subject = subject,
+      practitioner = null
+      /* TODO DO NOT pass this id to MeasureProcessor as this is treated as subject if subject is null.
```
pld, line 371: Interesting, we could only set `practitioner` when `subject` is NOT null, and note that in the docstring for this function.

But that still feels a little weird.

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/MeasureReportViewModel.kt
```diff
@@ -360,164 +327,188 @@ constructor(
...
+    subjectXFhirQuery: String?
+  ): List<MeasureReport> {
+    val measureReport = mutableListOf<MeasureReport>()
+    withContext(dispatcherProvider.io()) {
+      if (subjectXFhirQuery?.isNotEmpty() == true) {
+        fhirEngine
+          .search(subjectXFhirQuery)
+          .map {
+            // TODO a hack to prevent missing subject in case of Group based reports where
+            // MeasureEvaluator looks for Group members and skips the Group itself
```
pld, line 339: Does this happen because of the way a measure is written, or does evaluate-measure always look for a Group member whenever you pass it a Group?

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/MeasureReportViewModel.kt
```diff
@@ -360,164 +327,188 @@ constructor(
...
-      defaultRepository.addOrUpdate(resource = measureReport)
-      _measureReportPopulationResultList.addAll(
-        formatPopulationMeasureReport(measureReport, indicatorTitle)
-      )
+      measureReport.forEach { defaultRepository.addOrUpdate(resource = it) }
     }
+    return measureReport
+  }
+
+  fun runMeasureReport(
```
pld, line 357: could you please add a docstring

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/MeasureReportViewModel.kt
```diff
@@ -360,164 +327,188 @@ constructor(
...
-            )
+            }
+          }
+          .forEach { data.add(it) }
+      }
+
+    return data
+  }
+
+  fun formatSupplementalData(
```
pld, line 484: could you please add a docstring

pld (approved):
lgtm! @maimoonak a couple comments but nothing blocking, feel free to merge if this is ready from your perspective

pld (dismissed):
pls add the docstrings noted

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/MeasureReportViewModel.kt
maimoonak: Yeh. The measure evaluator ignores the given Group subject and just looks for embedded entities

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/MeasureReportViewModel.kt
```diff
@@ -282,24 +282,36 @@ constructor(
...
+                  retrievePreviouslyGeneratedMeasureReports<MeasureReport>(
+                    fhirEngine,
+                    startDateFormatted,
+                    endDateFormatted,
+                    config.url
+                  )
+
+                // if report is of current month or does not exist generate a new one and replace
+                // existing
+                if (startDateFormatted.contentEquals(
```
pld, line 296: Don't the start and the end date need to match?

pld (changes requested):
please address open comments

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/MeasureReportViewModel.kt
maimoonak: the start and end dates in reports are start or end of months.. so if start date is of current month or end date is falling in current month we need to regenerate the reports.

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/MeasureReportViewModel.kt
maimoonak: it is here. I do not know why it does not show it as outdated into main view

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/MeasureReportViewModel.kt
pld: weird looks like a git diff error

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/MeasureReportViewModel.kt
pld: OK got it, I had the condition reversed, but something still feels odd here. One thing to consider is that reports don't have to only be for a month.

If a report starts on the 5th of the month, and ends on the 15th of the month, we'd want to use the existing report as long as end date is in the past. (I'm assuming there's some logic elsewhere requiring that end date is after start date, can you check date please?)

Maybe this check is simpler, if the end date is before the current date and there is an existing report, load it, otherwise generate.

Can you add some tests around reports that are less than and greater than 1 month?

pld (dismissed):

pld (approved):

PR #1930, opened 2023-01-03
Ignore compose preview functions from codecov

### LINKED ISSUE #1929: Ignore compose preview functions from codecov
**Describe the Issue**
Compose `@Preview` annotation is used to quickly preview the Jetpack Compose UI that engineers have created during development. It's not included in production because ProGuard supposed to remove them.

So it's good to ignore them from codecov.

FikriMilano:

Fixes #1929 

**Checklist**


pld:
Could we override the @Compose decorator to do this, instead of adding an additional decorator?

FikriMilano:
not really sure, but it's worth trying

FikriMilano:
awesome, it works.

so I created a custom annotation, that includes both Preview and Exclude codecov annotation.
the naming must include `Generated` so the exclude functionality can work.

now we just need to use 1 annotation to preview and exclude :D
thx for the idea @pld

FikriMilano:
will update the rest on next week CHT

android/engine/src/main/java/org/smartregister/fhircore/engine/ui/components/PinView.kt
```diff
@@ -214,44 +214,44 @@ fun PinCell(
   }
 }
 
-@ExperimentalComposeUiApi
-@Preview
+@PreviewExcludeGenerated
```
ellykits, line 217: Is it possible to overload the `@Preview` annotation? Instead of creating a new one?, but again it will be confusing and it is easier to forget.

ellykits (commented):
LGTM. Have you updated every Preview annotation in the codebase?

android/engine/src/main/java/org/smartregister/fhircore/engine/ui/components/PinView.kt
FikriMilano: Yeah that would definitely be better, but kotlin annotation class doesn't allow to do that

FikriMilano: I tried make it inherit from @Preview but that doesn't work either

FikriMilano:
> LGTM. Have you updated every Preview annotation in the codebase?

Yeah, all of them

pld (approved):

PR #1934, opened 2023-01-04
bump coverage

owais-vd:

Fixes #[issue number] 

**Checklist**

### DIFF
--- a/android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
+++ b/android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
@@ -371,8 +371,9 @@ constructor(
     }
 
     /**
-     * This function takes a [attribute] as a field name and [type] as a source class to find out
-     * the actual field from full class hierarchy or return null in case of missing
+     * This function takes a [attribute] as a field name and [type] as a source class to find the
+     * actual field from full class hierarchy. If the field does not exist anywhere in the class
+     * hierarchy we return null.
      *
      * @return optional [Field]
      */
--- a/android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
+++ b/android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
@@ -21,7 +21,8 @@ import com.google.android.fhir.logicalId
 import dagger.hilt.android.qualifiers.ApplicationContext
 import java.lang.reflect.Field
 import java.text.SimpleDateFormat
-import java.util.*
+import java.util.Date
+import java.util.Locale
 import javax.inject.Inject
 import org.apache.commons.jexl3.JexlBuilder
 import org.apache.commons.jexl3.JexlException
@@ -359,8 +360,8 @@ constructor(
       list.filter { getValue(it, attribute) == attributeValue }
 
     /**
-     * This function takes a [resource] as source class and a [attribute] as field name
-     * to find out the actual value by using the java reflection API.
+     * This function takes a [resource] as source class and a [attribute] as field name to find out
+     * the actual value by using the java reflection API.
      *
      * @return optional [Any] primitive type value
      */
--- a/android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
+++ b/android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
@@ -19,6 +19,7 @@ package org.smartregister.fhircore.engine.rulesengine
 import android.content.Context
 import com.google.android.fhir.logicalId
 import dagger.hilt.android.qualifiers.ApplicationContext
+import java.lang.reflect.Field
 import java.text.SimpleDateFormat
 import java.util.Date
 import java.util.Locale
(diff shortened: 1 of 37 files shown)


android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
```diff
@@ -359,9 +360,19 @@ constructor(
...
-      return (property.get(resource) as? PrimitiveType<*>)?.value
+      val property = getDeclaredField(attribute, resource.javaClass)
+      property?.isAccessible = true
+      return (property?.get(resource) as? PrimitiveType<*>)?.value
+    }
+
+    private fun getDeclaredField(attribute: String, type: Class<*>): Field? {
+      return try {
+        type.getDeclaredField(attribute)
+      } catch (ex: Exception) {
```
pld, line 371: what's a narrower exception you can catch here?

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
```diff
@@ -359,9 +360,19 @@ constructor(
...
+        if (type.superclass != null) {
+          getDeclaredField(attribute, type.superclass)
+        } else null
```
pld, line 374: do we want to raise instead or returning null here?

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
```diff
@@ -359,9 +360,19 @@ constructor(
...
     private fun getValue(resource: Resource, attribute: String): Any? {
-      val property = (resource.javaClass).getDeclaredField(attribute)
-      property.isAccessible = true
-      return (property.get(resource) as? PrimitiveType<*>)?.value
+      val property = getDeclaredField(attribute, resource.javaClass)
+      property?.isAccessible = true
+      return (property?.get(resource) as? PrimitiveType<*>)?.value
+    }
+
+    private fun getDeclaredField(attribute: String, type: Class<*>): Field? {
```
pld, line 368: please add a docstring above

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
```diff
@@ -359,9 +360,19 @@ constructor(
       list.filter { getValue(it, attribute) == attributeValue }
 
     private fun getValue(resource: Resource, attribute: String): Any? {
```
pld, line 362: please add a docstring above

pld (changes requested):

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
owais-vd: done now catching `NoSuchFieldException`

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
owais-vd: returning `null` because might be a chance the field that we want is not available in the entire class hierarchy and `null` would be a case of empty value in the application.

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
owais-vd: done

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
owais-vd: done

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
pld: cool, great explanation, I moved that into a comment

pld (approved):

pld:
This is my fault, I'll run spotless...

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
```diff
@@ -19,9 +19,9 @@ package org.smartregister.fhircore.engine.rulesengine
 import android.content.Context
 import com.google.android.fhir.logicalId
 import dagger.hilt.android.qualifiers.ApplicationContext
+import java.lang.reflect.Field
 import java.text.SimpleDateFormat
-import java.util.Date
-import java.util.Locale
+import java.util.*
```
pld, line 24: please replace wildcard with specific libraries

pld (changes requested):

pld (approved):

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
ndegwamartin: @owais-vd would we want to check `getDeclaredField(attribute, type.superclass)` first and then if that throws the exception we catch it and then try the `type.getDeclaredField(attribute)`?

owais-vd: sorry, I'm not getting your point, why do we go superclass first? could you add more clarification? Thanks

ndegwamartin: oh I see, I thought the second parameter for getDeclaredField was to strongly type the value returned, such that if the field is a resource like patient and the type is Patient then we want to evaluate that first. If it fails then we fall back to the one without type.

ndegwamartin (approved):

PR #1987, opened 2023-01-24
Fix issue 1971 [Individual Profile view enhancement]

### LINKED ISSUE #1971: [UX] Individual Profile view enhancement
**Describe the enhancement**
**Screenshot 1**
image

1. Inappropriate Space - This space shouldn’t be there
3. Space  - Space between the 2 text items doesn’t match the designs’ spacing
4. Text Properties - The text properties for the labels and corresponding body content don’t match the design. The labels are smaller 13px and grey, the body is darker and bigger 14px.
5. Text Properties - The text properties for the heading don’t match with the designs. Plus the padding between heading and container below doesn’t match design
6. Container State and properties

- When the container is empty, we should inform the user that there no visits.
- the container containers a drop down shadow
**Adjusted Design**
image

**Screenshot 2**
image

1. The care plans tags are missing
2. The container of all visit buttons has different padding properties compared to design and missing drop shadow
3. Visit Buttons

- the buttons should be centered
- the content in buttons should be centered
- due date is missing
- the boarders are thicker and have more contrast/opacity as compared to the design
- the colors are different from design
- the text properties are different from design

**Adjusted Design**
image

SebaMutuku:

Fixes #1971  

**Checklist**


docs/configurations/application-config.mdx
```diff
@@ -45,6 +45,39 @@ There can only one instance of application configuration for the entire applicat
 }
 ```
 
+## Configuring widgets in FlexRow alignment
```
pld, line 48: 🎉

pld (approved):

docs/configurations/application-config.mdx
ellykits: @SebaMutuku Nice to include this in the documentation. But I think the documentation here should be for `LIST` view widget which should be isolated from application configurations. I would suggest we create a separate directory for widget configurations then include this inside the `ListProperties` documentation.

pld (approved):

PR #2070, opened 2023-02-21
Fix issue 2036 [Configuration-Filtering registers based on resources attributes ]

### LINKED ISSUE #2036: [Configuration] Filtering registers based on resources attributes
- When a CHSS logs into their application, they should only have access to data that is from the CHAs under them. They should also be able to view a list of only their CHAs on the CHA register. 

- [ ] When a CHSS opens the patient data registers from the CHA profile the records displayed should be filtered by the. CHA (Practitioner ID)
     - [ ] All the records created by the app are tagged by the `practitioner Id` on the `meta.tag` attribute.  
     - [ ] Write a filter that can accept expressions (probably) FHIR path to filter the resources displayed on the register by the clicked practitioner Id. 

- [This](https://github.com/opensrp/fhircore/issues/2001) work by @SebaMutuku filters resources in a LIST view by specific FHIR path expressions. I think we can extend it to filter all the registers.
 
N/A

**Additional context**
- A CHSS acts as a supervisor to the CHA. A CHSS is normally assigned 5-10 CHAs under them. CHSS is responsible for reviewing the CHA work and thus needs to see each CHA's updated data after P2P or a sync.

SebaMutuku:

Fixes #2036 

**Checklist**


android/engine/src/main/java/org/smartregister/fhircore/engine/data/local/register/RegisterRepository.kt
```diff
@@ -272,6 +288,33 @@ constructor(
     return newListRelatedResources ?: mutableListOf()
   }
 
+  /**
+   * This function returns a list of filtered resources. The required list is obtained from
+   * [relatedResourceMap], then a filter is applied based on the condition returned from the
+   * extraction of the [ListResource] conditional FHIR path expression
+   */
+  private fun filteredResourcesPerPractitioner(
```
pld, line 296:
```suggestion
  private fun filteredResources(
```

pld: and similar changes within this function

android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/view/ListProperties.kt
```diff
@@ -54,5 +55,12 @@ data class ListResource(
   val relatedResourceId: String? = null,
   val resourceType: ResourceType,
   val conditionalFhirPathExpression: String? = null,
+  val practitionerFhirPathExpression: String? = null,
   val relatedResources: List<ExtractedResource> = emptyList()
 )
+
+@Serializable
+data class PractitionerConfig(
```
pld, line 63:
```suggestion
data class FilterConfig(
```

android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/view/ListProperties.kt
```diff
@@ -40,7 +40,8 @@ data class ListProperties(
   val showDivider: Boolean = true,
   val emptyList: NoResultsConfig? = null,
   val orientation: ListOrientation = ListOrientation.VERTICAL,
-  val resources: List<ListResource> = emptyList()
+  val resources: List<ListResource> = emptyList(),
+  val practitionerList: List<PractitionerConfig> = emptyList()
```
pld, line 44:
```suggestion
  val filters: List<FilterConfig> = emptyList()
```

pld (changes requested):
we need to rework this not have anything specific to a Practioner in it. The goal of the issue is to allow filtering or an arbitrary resource by arbitrary attributes in that resource. Looking through a code it looks this is mainly a naming problem the architecture looks like it's on the correct path overall.

I put some ideas towards that as suggested changes, I'm not sure how practitionerId generalizes, or if it's acutally needed

android/quest/src/main/java/org/smartregister/fhircore/quest/util/extensions/ConfigExtensions.kt
```diff
@@ -71,7 +74,20 @@ fun List<ActionConfig>.handleClickEvent(
...
+                  .map {
+                    ActionParameter(
+                      key = it.key,
+                      paramType = it.paramType,
+                      dataType = it.dataType,
+                      linkId = it.linkId,
+                      value = it.value.interpolate(resourceData?.computedValuesMap ?: emptyMap())
+                    )
+                  }
+                  .toTypedArray()
```
pld, line 90: is this duplicating 49-60? Also maybe duplicating 125-139 in ProfileViewModel, https://github.com/opensrp/fhircore/pull/2070/files#diff-8d649a3643b25b19e29bc3d0bdc0583811450a85fd92e128e6a6615f8fae07fbR125

pld (changes requested):
awesome progress towards configurable filters! left some pretty straightforward comments, looks very close to merging.

one broader question, is there a name more specific than `params` we can use throughout? What exactly do these params represent?

android/engine/src/test/java/org/smartregister/fhircore/engine/data/local/register/RegisterRepositoryTest.kt
```diff
@@ -234,6 +238,92 @@ class RegisterRepositoryTest : RobolectricTest() {
     runBlocking {
       val recordsCount = registerRepository.countRegisterData("patientRegister")
 
+      Assert.assertEquals(20, recordsCount)
+    }
+  }
+  @Test
```
pld, line 244:
```suggestion
  }
  
  @Test
```

pld (approved):

PR #2085, opened 2023-02-24
App is force closing while trying to log in while offline

owais-vd:

Fixes https://github.com/opensrp/fhir-resources/issues/868

**Checklist**

### DIFF
--- a/android/quest/src/main/java/org/smartregister/fhircore/quest/ui/login/AccountAuthenticator.kt
+++ b/android/quest/src/main/java/org/smartregister/fhircore/quest/ui/login/AccountAuthenticator.kt
@@ -25,11 +25,13 @@ import android.content.Intent
 import android.os.Bundle
 import androidx.core.os.bundleOf
 import dagger.hilt.android.qualifiers.ApplicationContext
+import java.net.UnknownHostException
 import java.util.Locale
 import javax.inject.Inject
 import org.smartregister.fhircore.engine.data.remote.shared.TokenAuthenticator
 import org.smartregister.fhircore.engine.data.remote.shared.TokenAuthenticator.Companion.AUTH_TOKEN_TYPE
 import retrofit2.HttpException
+import timber.log.Timber
 
 class AccountAuthenticator
 @Inject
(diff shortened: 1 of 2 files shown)


android/quest/src/main/java/org/smartregister/fhircore/quest/ui/login/AccountAuthenticator.kt
```diff
@@ -78,8 +80,12 @@ constructor(
         authToken =
           try {
             tokenAuthenticator.refreshToken(refreshToken)
-          } catch (httpException: HttpException) {
-            ""
+          } catch (ex: Exception) {
```
dubdabasoduba, line 83: @owais-vd have you tried the unknown host exception?

dubdabasoduba (changes requested):

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/login/AccountAuthenticator.kt
owais-vd: yes tired and got `UnknownHostException`

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/login/AccountAuthenticator.kt
```diff
@@ -78,8 +80,12 @@ constructor(
...
+            Timber.e(ex)
+            when (ex) {
+              is HttpException, is UnknownHostException -> ""
+              else -> throw ex
```
allan-on, line 87: Would this be handled elsewhere? 
Looks like there's potential for a crash if we rethrow an Exception. 🤔

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/login/AccountAuthenticator.kt
pld: k can we update to catch both HttpException and UnknownHostException? We'd like to avoid using a generic

pld: oh sorry! I misunderstood and code and didn't realize this is a strategy for multicatch in kotlin, this makes sense!

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/login/AccountAuthenticator.kt
pld: true but I think that's ok, that not this functions responsibility

pld (approved):
lgtm!

pld:
looks good in app, great work on the tests!

PR #2086, opened 2023-02-24
ordering of registers by related resources last updated

### LINKED ISSUE #2035: [Configuration] Ordering of registers by related resources last updated
Ordering of registers by last updated resource. 
At the moment, when a new family is registered it shows at the top of the register - this implementation is using the group resource. However, there is a need to enhance this so that the family's last update will appear at the top. This will also apply to the commodity register. 

**Describe the enhancement**
- On the registers, order the list by the last updated resource [relatedresources]. 
- On the household register, if any member of the family record is updated, this family will appear at the top of the family register. 
- On any other register, the client with the most recently updated record will appear at the top of the list, the rest will be organized in chronological order from the most recent to the furthest updated. 
- On the stock/inventory register, the commodity that was adjusted/ restocked last will appear at the top of the list, and the rest will be on order of the latest updated to the furthest.  

**Acceptance criteria**
- [ ] The household register shows the most recently updated family using [relatedresources] - the family with clients with the most recently updated will be at the top of the register list. 
- [ ] In other registers, the client with the most recently updated record is shown at the top of the register list.
- [ ] Stock/inventory register, the commodity with the most recently updated physical count [adjustment/restock] is shown at the top of the register list.

aurangzaibumer:

Fixes #2035

**Checklist**


android/quest/src/main/java/org/smartregister/fhircore/quest/ui/main/AppMainViewModel.kt
```diff
@@ -306,6 +313,20 @@ constructor(
...
 
   suspend fun onQuestionnaireSubmission(questionnaireSubmission: QuestionnaireSubmission) {
+    // Patient/1423423423
+    questionnaireSubmission.questionnaireResponse.subject?.run {
+      val resource: Resource =
+        defaultRepository.loadResource(questionnaireSubmission.questionnaireResponse.subject)
+      defaultRepository.addOrUpdate(resource = resource)
+    }
+    // update group
+    questionnaireSubmission.questionnaireConfig.groupResource?.groupIdentifier.let {
```
aurangzaibumer, line 323: @ahsanbhatti49

aurangzaibumer:
I've completed task on the parameters passing between screens. you can review and provide feedback so I can continue on unit test on this @ellykits. 

Meanwhile @ahsanbhatti49  is working on lastUpdate task onformsubmission.

thanks.
cc @f-odhiambo

dubdabasoduba:
The current implementation only seems to work if we update or add a patient.  

- Right now we seem to have updated [this](https://github.com/opensrp/fhircore/blob/76e55b995ec7ffbf99167c7042d0009d46fb6533/android/quest/src/main/java/org/smartregister/fhircore/quest/ui/main/AppMainViewModel.kt#L315-L341) piece of code to do the updates. I don't think this will work with main resources that are not Groups. 
- This won't also work when the Tasks are performed on the patient profile. 

**How I expected the feature to work.** 
- [ ] Ability to pass the id of the main resource from the register to the different profiles e.g pass the family Id up to the patient profile. Pass the Group Id for commodities to the commodity profile. 
- [ ] Load these Ids to the questionnaires opened from these profiles
- [ ] Use the ids in the StructureMaps to update the main resources `lastUpdated`. 


https://user-images.githubusercontent.com/4817531/222030255-1034eb91-7bd1-4c8d-9ec7-4b13ea8fdb1b.mp4

ahsanbhatti49:
> The current implementation only seems to work if we update or add a patient.
> 
> * Right now we seem to have updated [this](https://github.com/opensrp/fhircore/blob/76e55b995ec7ffbf99167c7042d0009d46fb6533/android/quest/src/main/java/org/smartregister/fhircore/quest/ui/main/AppMainViewModel.kt#L315-L341) piece of code to do the updates. I don't think this will work with main resources that are not Groups.
> * This won't also work when the Tasks are performed on the patient profile.
> 
> **How I expected the feature to work.**
> 
> * [ ]  Ability to pass the id of the main resource from the register to the different profiles e.g pass the family Id up to the patient profile. Pass the Group Id for commodities to the commodity profile.
> * [ ]  Load these Ids to the questionnaires opened from these profiles
> * [ ]  Use the ids in the StructureMaps to update the main resources `lastUpdated`.

@dubdabasoduba  As per my discussion with @maimoonak

ahsanbhatti49:
@dubdabasoduba   As per my discussion with @maimoonak , We dont need to explicitly update any resource when ever there is any update . When ever there is any update in the resource its lastupdate is already updated, And if it is associated with any group, then we should update it group resourvce as well otherwise we dont have to explicitly update its _lastupdate.

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireViewModel.kt
```diff
@@ -110,6 +111,21 @@ constructor(
       ?.extractLogicalIdUuid()
   }
 
+  private fun readFileFromAssets(filename: String, application: Application): String {
+    return application.assets.open(filename).bufferedReader().use { it.readText() }
+  }
+  //  suspend fun loadQuestionnaire(
```
dubdabasoduba, line 117: Remove this piece of code

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/profile/ProfileViewModel.kt
```diff
@@ -155,10 +156,15 @@ constructor(
...
                       )
+                    var groupResourceId: String? = null
+
                     val params =
                       actionConfig
                         .params
                         .map {
+                          if (it.key == "familyLogicalId") {
+                            groupResourceId = it.value
+                          }
```
dubdabasoduba, line 167: This means the config authors have to name the rules a certain way. I don't think this is the correct implementation. 
- We can't control how people name their rules while authoring configs.

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/main/AppMainViewModel.kt
```diff
@@ -311,13 +306,31 @@ constructor(
   }
 
   suspend fun onQuestionnaireSubmission(questionnaireSubmission: QuestionnaireSubmission) {
+    // Patient/1423423423
+    questionnaireSubmission.questionnaireResponse.subject?.run {
+      val resource: Resource =
+          defaultRepository.loadResource(questionnaireSubmission.questionnaireResponse.subject)
+      defaultRepository.addOrUpdate(resource = resource)
+    }
+    // update group
```
dubdabasoduba, line 315: Are we still using the `groupResoruce` even after deciding to use the `questionnaire.action.params`?

dubdabasoduba (changes requested):

android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/QuestionnaireConfig.kt
```diff
@@ -35,7 +35,7 @@ data class QuestionnaireConfig(
   val resourceIdentifier: String? = null,
   val resourceType: String? = null,
   val confirmationDialog: ConfirmationDialog? = null,
-  val groupResource: GroupResourceConfig? = null,
+  var groupResource: GroupResourceConfig? = null,
```
ellykits, line 38: Why are we converting this to mutable property?

ellykits: The config data models are majorly used to read and hold configuration data and thus should not mutate

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/profile/ProfileViewModel.kt
ellykits: Why is this hardcoded?

ellykits (changes requested):

ellykits:
@ahsanbhatti49 I think you need to refactor the implementation as we had discussed previously. The current solution is not generic.

android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/QuestionnaireConfig.kt
```diff
@@ -50,7 +50,7 @@ data class ConfirmationDialog(
 
 @Serializable
 data class GroupResourceConfig(
-  val groupIdentifier: String,
+  var groupIdentifier: String,
```
ellykits, line 53: Revert this to `val`

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireActivity.kt
```diff
@@ -106,6 +107,13 @@ open class QuestionnaireActivity : BaseMultiLanguageActivity(), View.OnClickList
           !it.value.isNullOrEmpty() &&
           !it.value.contains(STRING_INTERPOLATION_PREFIX)
       }
+    paramDataActionParams =
+      actionParams.filter {
+        it.paramType == ActionParameterType.PARAMDATA &&
+          !it.value.isNullOrEmpty() &&
+          !it.value.contains(STRING_INTERPOLATION_PREFIX)
```
ellykits, line 114: I do not think this check is necessary. Can we use a different `ActionParameterType` for this implementation? This will avoid ambiguity.

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireActivity.kt
```diff
@@ -106,6 +107,13 @@ open class QuestionnaireActivity : BaseMultiLanguageActivity(), View.OnClickList
...
+      }
+    questionnaireViewModel.setParamData(paramDataActionParams)
```
ellykits, line 116: I think the parameters are pre-loaded and interpolated when the questionnaire is launched. Why do we need to set new params?

ellykits: see the `launchQuestionnaire` function. It accepts a list of params that is provided by the caller.

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireViewModel.kt
```diff
@@ -407,8 +412,17 @@ constructor(
       )
       return
     }
-
-    defaultRepository.addOrUpdate(resource = questionnaireResponse)
+    defaultRepository.addOrUpdate(resource = questionnaireResponse) // update resource
+    // update the group/Or any other resource which is associated with this resource
+
+    if (paramDataActionParameter.isNotEmpty() && paramDataActionParameter[0].value.contains("/")) {
```
ellykits, line 418: If we use a distinct `ActionParameterType`, we will not need this check plus this assumption is not totally correct. We might have other parameter types that contain `/` that is not of our interest.

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireActivity.kt
ellykits: You can for instance use `ActionParameterType.UPDATABLE_IDS` for lack of a better name

ellykits (changes requested):

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireActivity.kt
ahsanbhatti49: launchQuestionnaire function accepts PREPOPULATED action params only. 
thats why i have initialized separate action params for PARAMDATA type.

pld (commented):
it's unclear 1) why we need a new action parameter type, 2) why there's any reference to familyLogicalId in the code

the idea of the issue is that the ordering can be linked to any specified date/datetime filed in any related resource, like in Elly's comment here, https://github.com/opensrp/fhircore/issues/2035#issuecomment-1430928769

android/engine/src/main/java/org/smartregister/fhircore/engine/domain/model/ActionParameterType.kt
```diff
@@ -22,5 +22,6 @@ import kotlinx.serialization.json.JsonNames
 enum class ActionParameterType {
   /** Represents parameters that are used to pre-populate Questionnaire items with initial values */
   @JsonNames("pre_populate", "PrePopulate") PREPOPULATE,
-  @JsonNames("param_data", "ParamData") PARAMDATA
+  @JsonNames("param_data", "ParamData") PARAMDATA,
+  @JsonNames("updateable_ids", "Updateable_Ids") UPDATEABLEIDS
```
ellykits, line 26:
```suggestion
  @JsonNames("updateable_ids", "UpdateableIds") UPDATEABLEIDS
```

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireActivity.kt
```diff
@@ -106,6 +106,9 @@ open class QuestionnaireActivity : BaseMultiLanguageActivity(), View.OnClickList
           !it.value.isNullOrEmpty() &&
           !it.value.contains(STRING_INTERPOLATION_PREFIX)
       }
+    questionnaireViewModel.setFamilyLogicalIdActionParam(
+      actionParams.firstOrNull { it.paramType == ActionParameterType.UPDATEABLEIDS }
```
ellykits, line 110: I do not think we need this implementation. Update the filter logic for action Parameters above to include `ActionParameterType.UPDATEABLEIDS` instead.

ellykits (changes requested):

android/engine/src/main/java/org/smartregister/fhircore/engine/domain/model/ActionParameterType.kt
pld: I'm not sure how I feel about the name "UPDATEABLEIDS" is there something more descriptive that we can choose?

From my understanding of the meaning, what about "UPDATE_DATE_ON_EDIT"?

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireViewModel.kt
```diff
@@ -109,6 +109,13 @@ constructor(
       .read(SharedPreferenceKey.PRACTITIONER_ID.name, null)
       ?.extractLogicalIdUuid()
   }
+  private var familyLogicalIdActionParam: ActionParameter? = null
```
pld: I don't think we want to use `familyLogicalId` anywhere, that's an arbitrary name in a config, this applies to `familyLogicalId` used throughout the code. I'm not entirely certain, but I think the abstraction is that we want a list of `updateableOnEditIds` that collects all the values where the paramType is `UPDATEABLEIDS` (or whatever we name it) cc @ellykits

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireViewModel.kt
```diff
@@ -109,13 +110,7 @@ constructor(
...
       ?.extractLogicalIdUuid()
   }
-  private var familyLogicalIdActionParam: ActionParameter? = null
-
-  fun setFamilyLogicalIdActionParam(params: ActionParameter?) {
-    if (params != null) {
-      this.familyLogicalIdActionParam = params
-    }
-  }
+  private var actionParamUpdatableId: ActionParameter? = null
```
pld, line 113: 👍

android/engine/src/main/java/org/smartregister/fhircore/engine/domain/model/ActionParameterType.kt
ellykits: This `UPDATE_DATE_ON_EDIT` sound better.

pld:
```suggestion
  @JsonNames("update_date_on_edit", "UpdateDateOnEdit") UPDATE_DATE_ON_EDIT
```

android/quest/src/main/assets/configs/app/profiles/default_profile_config.json
```diff
@@ -3145,6 +3144,13 @@
             "resourceIdentifier": "@{patientId}",
             "planDefinitions": [
               "9fac39bd-9750-4c1d-b355-1961de4e37f3"
+            ],
+            "params": [
+              {
+                "key": "familyLogicalId",
+                "value": "@{familyLogicalId}",
+                "paramType": "UPDATEABLEIDS"
```
pld, line 3152: please update to match changed value name

android/quest/src/main/assets/configs/app/registers/practitioner_register_config.json
```diff
@@ -117,7 +117,8 @@
                   }
                 }
 
-              ]
+              ],
+
```
pld, line 121: please revert, we don't use terminal commas in other places. can you else delete the empty line on 119, let's cleanup to remove any existing extra line breaks, thank you
```suggestion
              ]
```

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireViewModel.kt
```diff
@@ -407,8 +412,21 @@ constructor(
       )
       return
     }
-
-    defaultRepository.addOrUpdate(resource = questionnaireResponse)
+    defaultRepository.addOrUpdate(resource = questionnaireResponse) // update resource
+    // update the group/Or any other resource which is associated with this resource
```
pld, line 416: thank you for the comments! I know this is cleanup for previous work, but can you also please add a docstring to the `loadQuestionnaire` function itself explaining this function and the parameters, https://kotlinlang.org/docs/kotlin-doc.html#kdoc-syntax, thank you

android/quest/src/main/java/org/smartregister/fhircore/quest/util/IterableUtil.kt
```diff
@@ -28,3 +31,26 @@ package org.smartregister.fhircore.quest.util
...
+      if ((param.paramType == ActionParameterType.PARAMDATA) ||
+          (param.paramType == ActionParameterType.UPDATEABLEIDS) && !param.value.isNullOrEmpty()
+      ) {
+        paramsMap[param.key] = param.value
+      }
+    }
+    return paramsMap.toMap()
+  }
+  return emptyMap()
+}
```
pld, line 56: This looks like it should extend this function, https://github.com/opensrp/fhircore/blob/main/android/quest/src/main/java/org/smartregister/fhircore/quest/util/extensions/ConfigExtensions.kt#L140, not be added here

```suggestion
```

pld: in that extension you can use

```kotlin
param.paramType in listOf(ActionParameterType.PARAMDATA, ..)
```

to reduce duplication

pld (changes requested):
very close!

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/profile/ProfileViewModel.kt
pld, line 1: 👍

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireViewModel.kt
```diff
@@ -399,6 +406,10 @@ constructor(
       }
   }
 
+  /**
+   * update questionnaireResponse update Associated Resource by extracting logicalUuid from
+   * actionParamUpdatableId
```
pld, line 411:
```suggestion
   * Add or update the [questionnaireResponse] resource with the passed content, and if an [actionParamUpdatableId] is set also
   * update the resource it refers to by extracting its logicalIdUuid.
   *
   * @param questionnaire the [Questionnaire] this response is related to
   * @param questionnaireResponse the questionnaireResponse resource to save
```

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireViewModel.kt
```diff
@@ -409,8 +420,20 @@ constructor(
...
+      val resource =
+        defaultRepository.loadResource(
+          actionParamUpdatableId!!.value.extractLogicalIdUuid(),
+          actionParamUpdatableId!!
+            .value
+            .substringBefore("/")
+            .resourceClassType()
+            .newInstance()
+            .resourceType
+        )
```
pld, line 434: haven't we already verified it's not null? might also use a temp var

```suggestion
      val resource = actionParamUpdatableId.value.let {
        defaultRepository.loadResource(
          it.extractLogicalIdUuid(),
          it
            .substringBefore("/")
            .resourceClassType()
            .newInstance()
            .resourceType
        )
      }
```

pld (dismissed):

pld (approved):

pld (approved):

PR #2091, opened 2023-02-25
Improve performance when executing rules

### LINKED ISSUE #2016: [Rules Evaluation] Fix the slow evaluation of rules on the configs
**Describe the bug**
- Once we add multiple rules to the different view configs the pages seem to slow down considerably. 
- This is caused by the slow speed at which the rules are executed. 

**To Reproduce**
1. log in to the latest eCBIS release found [here](https://github.com/opensrp/fhircore/releases/tag/v0.1.9-ecbis-rc3). 
2. Click on any family, then click on a family member with generated tasks. 
3. You will notice the profile takes a bit of time to load and even has a blank screen displayed. 

**Expected behaviour**
- [ ] The screen should not take long to load. It should efficiently evaluate the rules and return the results to the screen for display. 

**Screenshots**
[blank-profile-page.webm](https://user-images.githubusercontent.com/4817531/217105113-20a163c1-d2e8-4c01-9301-3f5e266a7deb.webm)

ellykits:

Fixes #2016 

**Checklist**


android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesExecutor.kt
```diff
@@ -0,0 +1,208 @@
...
+    computedValuesMap: Map<String, Any>
+  ) =
+    this.map { resource ->
+      val listItemRelatedResources: Map<String, List<Resource>> =
+        listRelatedResources.associate { (id, resourceType, fhirPathExpression) ->
+          (id
+            ?: resourceType.name) to
+            rulesFactory.rulesEngineService.retrieveRelatedResources(
+              resource = resource,
+              relatedResourceKey = id ?: resourceType.name,
```
pld, line 103: so the constraint here is that you have to pass an ID or you can't have multiple listRelatedResources with the same resourceType. I think it would be nice to throw an error here if we see a id or resourceType.name that is duplicated

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesExecutor.kt
```diff
@@ -0,0 +1,208 @@
...
+ * config Id ( or resource type if the id is not configured) as key and value as list of [Resource]
+ * s in the map.
+ */
+fun LinkedList<RepositoryResourceData>.createRelatedResourcesMap():
+  MutableMap<String, MutableList<Resource>> {
+  val relatedResourcesMap = mutableMapOf<String, MutableList<Resource>>()
+  while (this.isNotEmpty()) {
+    val relatedResourceData = this.removeFirst()
+    relatedResourcesMap
+      .getOrPut(relatedResourceData.configId ?: relatedResourceData.resource.resourceType.name) {
```
pld, line 177: kinda the same issue as on line 103

pld: this is interesting, not sure how expensive each iteration in this loop is, if it's expensive enough worth considering using [parallelStream](https://blogs.oracle.com/javamagazine/post/java-parallel-streams-performance-benchmark) here. You would have to restructure to have it be tiered, like parallel process one level, to build next level, then parallel process next level etc, but that should be doable

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesExecutor.kt
```diff
@@ -0,0 +1,208 @@
...
+      relatedResourcesMap = relatedResourcesMap
+    )
+
+  private suspend fun List<Resource>.mapToResourceData(
+    relatedResourcesMap: MutableMap<String, MutableList<Resource>>,
+    ruleConfigs: List<RuleConfig>,
+    listRelatedResources: List<ExtractedResource>,
+    computedValuesMap: Map<String, Any>
+  ) =
+    this.map { resource ->
```
pld, line 96: also not sure how expensive these loop iterations are, you can use parallelStream with the entrySet if that gives you a speed up

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
```diff
@@ -114,44 +119,46 @@ constructor(
...
+          put(FHIR_PATH, fhirPathDataExtractor)
+          put(DATA, mutableMapOf<String, Any>())
+          put(SERVICE, rulesEngineService)
+          if (baseResource != null) {
+            put(baseResource.resourceType.name, baseResource)
+          }
+          relatedResourcesMap.forEach { put(it.key, it.value) }
+        }
+      val timeToFireRules = measureTimeMillis { rulesEngine.fire(rules, facts) }
+      Timber.d("Rule executed in $timeToFireRules millisecond(s)")
```
pld, line 141: do we want to wrap this in a feature flag
```suggestion
      if (TRACK_TIMING) {
        val timeToFireRules = measureTimeMillis { rulesEngine.fire(rules, facts) }
        Timber.d("Rule executed in $timeToFireRules millisecond(s)")
      } else {
        rulesEngine.fire(rules, facts)
      }
```
not sure, but I assume on compilation it will remove the conditional if TRACK_TIMING is a constant

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
```diff
@@ -114,44 +119,46 @@ constructor(
...
+  }
 
-    relatedResourcesMap.forEach { facts.put(it.key, it.value) }
-    rulesEngine.fire(Rules(customRules), facts)
-
-    return mutableMapOf<String, Any>().apply { putAll(computedValuesMap) }
+  fun generateRules(ruleConfigs: List<RuleConfig>): Rules {
+    val customRules =
+      ruleConfigs
+        .map { ruleConfig ->
```
pld, line 149: also worth checking parallelizing this

pld:
I guess in kotlin coroutines are preferred over parallel stream unless it's a very long list, and our lists aren't going to be that long, this looks like a useful parallel map definition we could drop in, https://stackoverflow.com/questions/34697828/parallel-operations-on-kotlin-collections/45794062#45794062

android/engine/src/main/java/org/smartregister/fhircore/engine/data/local/register/RegisterRepository.kt
```diff
@@ -113,8 +103,8 @@ constructor(
 
     // Retrieve data for each of the configured related resources
     // Also retrieve data for nested related resources for each of the related resource
-    baseResources.map { baseResource: Resource ->
-      val currentRelatedResources = LinkedList<RelatedResourceData>()
+    return baseResources.map { baseResource: Resource ->
+      val currentRelatedResources = LinkedList<RepositoryResourceData>()
 
       relatedResourcesConfig.forEach { resourceConfig: ResourceConfig ->
```
pld, line 109: is the `searchRelatedResources` expensive? This might be another loop to parallelize

pld (commented):
This isn't a line you touched but lines 162-164 of android/engine/src/main/java/org/smartregister/fhircore/engine/data/local/register/RegisterRepository.kt look like they're performing three sequential maps (more or less) that we could combine into a single map or for loop over the elements of `fhirPathDataExtractor.extractData(baseResource, fhirPathExpression)`

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
ellykits: We can check if `BuildConfig.Debug` is true instead, this log only makes sense while debugging.

android/engine/build.gradle
```diff
@@ -131,8 +131,6 @@ dependencies {
     // CQL
     implementation group: 'xerces', name: 'xercesImpl', version: '2.12.2'
     implementation(group: "com.github.java-json-tools", name: "msg-simple", version: "1.2");
-   // implementation 'org.codehaus.woodstox:woodstox-core-asl:4.4.1'
```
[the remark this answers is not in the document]
ellykits: Yes this can be removed.

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/MeasureReportViewModel.kt
```diff
@@ -407,24 +421,26 @@ constructor(
...
                 )
             }
           }
 
-        val indicators =
-          it.value.flatMap { report ->
+        val theIndicators =
+          entry.value.flatMap { report ->
             val formatted = formatSupplementalData(report.contained, report.type)
             val title = indicators.find { it.url == report.measure }?.title ?: ""
```
pld, line 445: this is probably expensive, especially since it's in a nested loop.

how about we create a map from indicator.url to indicator.title? so we can do this in constant time.

so somewhere outside both these loops, 
```kotlin
indicatorUrlToTitleMap = Map...
```

then here

```suggestion
            val title = indicatorUrlToTitleMap.getOrDefault(report.measure, "")
```

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/MeasureReportViewModel.kt
```diff
@@ -450,14 +466,16 @@ constructor(
...
           )
         )
       }
 
     measureReports
-      .takeIf { it.all { it.type != MeasureReport.MeasureReportType.INDIVIDUAL } }
+      .takeIf {
+        it.all { measureReport -> measureReport.type != MeasureReport.MeasureReportType.INDIVIDUAL }
+      }
       ?.forEach { report ->
```
pld, line 479: I think we can avoid a loop by pushing the takeIf into the forEach as a conditional, if this is something kotlin will optimize on compile time, then nevermind, but if not this should save time

```suggestion
      .forEach { report ->
        if (report.type == MeasureReport.MeasureReportType.INDIVIDUAL) return@forEach
```

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/MeasureReportViewModel.kt
```diff
@@ -473,21 +491,21 @@ constructor(
...
+                  )
                 }
           }
           .mapNotNull {
             it.first.findPopulation(MeasurePopulationType.NUMERATOR)?.let { count ->
               MeasureReportPopulationResult(
                 title = it.first.id.replace("-", " "),
-                indicatorTitle = indicators.find { it.url == report.measure }?.title ?: "",
+                indicatorTitle =
+                  indicators.find { reportConfig -> reportConfig.url == report.measure }?.title
```
pld, line 507: this again should use the map we build

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/MeasureReportViewModel.kt
```diff
@@ -504,28 +522,29 @@ constructor(
...
-      .let {
+      .let { observations ->
         // the observations would have key as coding.code=populationId and value as codeableConcept
-        it
+        observations
           .groupBy { it.codingOf(POPULATION_OBS_URL)?.display }
           .filter { it.key.isNullOrBlank().not() }
-          .map {
-            it.key!! to
+          .map { entry ->
```
pld, line 541: can group this filter and map to reduce looping

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/MeasureReportViewModel.kt
```diff
@@ -473,21 +491,21 @@ constructor(
                 .flatMap { it.stratum }
                 .filter { it.hasValue() && it.value.hasText() }
                 .map { stratifier ->
```
pld, line 493: can we replace the 3 implicit loops above with 1 explicit loop and conditionals?

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesExecutor.kt
ndegwamartin: I'm guessing this is so that we warn the user that they misconfigured with at duplicate?? If so I think an exception would be most useful to the writer of the configs, otherwise by the time the exception shows on the app it is probably too late. 

Once we implement a configs validator such a case should not happen and the issue would have been caught before hand

If we still need to throw/log such an exception let me know and we can refactor to accommodate this

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesExecutor.kt
ndegwamartin: Same as above on _Duplicate Exceptions_

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesExecutor.kt
pld: yea exactly, I think it's better to throw here and catch with a message somewhere higher up the call stack, even if we validate configs we're going to run this against arbitrary FHIR APIs (eventually) so we can't assume the config is valid

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesExecutor.kt
pld: yea let's throw, or maybe this calls for a helper that we call in both places, the code looks similar, if it's the same a helper makes sense that also handles the throwing

ndegwamartin:
@pld here are some resources on concurrency/parallelism on kotlin:
- https://kotlinlang.org/docs/multiplatform-mobile-concurrency-and-coroutines.html#platform-concurrency
- https://www.baeldung.com/kotlin/threads-coroutines
- https://betterprogramming.pub/the-difference-between-concurrency-and-parallelism-explained-using-kotlin-83f4159581d
- https://medium.com/android-news/kotlin-coroutines-threads-concurrency-and-parallelism-101-78a56e09d373
- https://jivimberg.io/blog/2018/05/04/parallel-map-in-kotlin/

pld:
> @pld here are some resources on concurrency/parallelism on kotlin:
> 
>     * https://kotlinlang.org/docs/multiplatform-mobile-concurrency-and-coroutines.html#platform-concurrency
> 
>     * https://www.baeldung.com/kotlin/threads-coroutines
> 
>     * https://betterprogramming.pub/the-difference-between-concurrency-and-parallelism-explained-using-kotlin-83f4159581d
> 
>     * https://medium.com/android-news/kotlin-coroutines-threads-concurrency-and-parallelism-101-78a56e09d373
> 
>     * https://jivimberg.io/blog/2018/05/04/parallel-map-in-kotlin/

I added a pmap function with an example usage in some of places I mentioned, https://github.com/opensrp/fhircore/pull/2091/commits/3a123d8bcf71527070b0400954649d518538058e

Probably doing more than low hanging fruit of replacing obvious non-interactive loops with a pmap is outside scope of this PR, for anything not obviously parallel please move to a new issue and we can addresses parallelization more there.

Cut scope where you need to get something we can merge by the 3rd.

android/engine/src/test/java/org/smartregister/fhircore/engine/rulesengine/RulesFactoryTest.kt
```diff
@@ -63,21 +64,46 @@ import org.smartregister.fhircore.engine.app.fakes.Faker
 import org.smartregister.fhircore.engine.configuration.ConfigurationRegistry
 import org.smartregister.fhircore.engine.domain.model.RuleConfig
 import org.smartregister.fhircore.engine.robolectric.RobolectricTest
+import org.smartregister.fhircore.engine.rule.CoroutineTestRule
 import org.smartregister.fhircore.engine.util.fhirpath.FhirPathDataExtractor
 
+fun populateTestPatient(): Patient {
```
pld, line 70: obviously needs to change, is this better placed in a new helper file?

android/engine/src/main/java/org/smartregister/fhircore/engine/data/local/register/RegisterRepository.kt
pld: this is already parallelized

pld: my bad, I missed that

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/profile/ProfileViewModel.kt
```diff
@@ -64,35 +68,55 @@ constructor(
...
         ProfileUiState(
           resourceData = resourceData,
-          profileConfiguration = retrieveProfileConfiguration(profileId),
-          snackBarTheme = applicationConfiguration.snackBarTheme
+          profileConfiguration = profileConfigs,
+          snackBarTheme = applicationConfiguration.snackBarTheme,
+          showDataLoadProgressIndicator = false
         )
+
+      profileConfigs.views.retrieveListProperties().forEach {
```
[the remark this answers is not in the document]
pld: this probably needs to go to an issue, I'd be scared to in parallel update a data structure, if this loop is expensive enough to parallelize, we use pmap to create a list of [id, listResourceData], then iterate that list to update `listResourceDataMapState`

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesExecutor.kt
pld: resolving since this will get thrown by retrieveRelatedResources

pld (approved):
as @ekigamba pointed out and @ndegwamartin verified, the loop optimizations I'm calling out here don't seem to be having much of an impact, approving this for merge, we can handle the optimizations in order of impact expected based on testing

pld (dismissed):

pld (approved):

PR #2118, opened 2023-03-04
test removing some lines

pld:


codecov.yml
```diff
@@ -1,34 +1,39 @@
...
-      - "/home/runner/work/fhircore/fhircore/android/engine"
+    - ^/home/runner/work/fhircore/fhircore/android/engine.*
   geowidget:
     paths:
-      - "/home/runner/work/fhircore/fhircore/android/geowidget"
+    - ^/home/runner/work/fhircore/fhircore/android/geowidget.*
   quest:
     paths:
-      - "/home/runner/work/fhircore/fhircore/android/quest"
+    - ^/home/runner/work/fhircore/fhircore/android/quest.*
```
ndegwamartin, line 37: Should these path definitions be starting with a caret?

pld: yea maybe not, I copied this directly from here, which was the last valid run, https://app.codecov.io/github/opensrp/fhircore/commit/5182221f688cfb24ae13720bb50221c2e040107e

codecov.yml
```diff
@@ -1,34 +1,39 @@
 codecov:
-  disable_default_path_fixes: yes
+  disable_default_path_fixes: true
```
ndegwamartin, line 2: Also since we've disabled the default path fixes we probably need to specify our paths manually

pld: let me try removing this line, this has always been disabled, but worth a try

pld:
yay this hash leads to processes the coverage reports successfully, https://github.com/opensrp/fhircore/pull/2118/commits/d803c43d5f5a33e3cf6eba29b8cfd1d6aa01b3aa, remaining problems are
- status is being reported for each flag, we don't care about per flag status

PR #2120, opened 2023-03-06
Implement Concurrency For Performance Optimization : RnD

ndegwamartin:
Notes

- Optimizations to Profile page loading
- Optimizations to Register page

**Checklist**

### DIFF
--- a/android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesExecutor.kt
+++ b/android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesExecutor.kt
@@ -18,6 +18,7 @@ package org.smartregister.fhircore.engine.rulesengine
 
 import com.google.android.fhir.logicalId
 import java.util.LinkedList
+import java.util.concurrent.ConcurrentLinkedQueue
 import javax.inject.Inject
 import org.hl7.fhir.r4.model.Resource
 import org.smartregister.fhircore.engine.configuration.view.CardViewProperties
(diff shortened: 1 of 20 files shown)


android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesExecutor.kt
```diff
@@ -80,26 +81,24 @@ class RulesExecutor @Inject constructor(val rulesFactory: RulesFactory) {
...
   private suspend fun List<Resource>.mapToResourceData(
     relatedResourcesMap: MutableMap<String, MutableList<Resource>>,
     ruleConfigs: List<RuleConfig>,
-    ruleConfigsKey: String,
     listRelatedResources: List<ExtractedResource>,
     computedValuesMap: Map<String, Any>
   ) =
     this.map { resource ->
-      val start = System.currentTimeMillis()
+      val start0 = System.currentTimeMillis()
```
ellykits, line 101: You can use Kotlin's lambda function [measureTimeMillis](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.system/measure-time-millis.html)

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesExecutor.kt
```diff
@@ -111,16 +110,20 @@ class RulesExecutor @Inject constructor(val rulesFactory: RulesFactory) {
               relatedResourcesMap = relatedResourcesMap
             )
         }
+      Timber.d(
+        "ListRelatedResources.associate executed in ${ System.currentTimeMillis() - start0 } millisecond(s)"
+      )
 
-      val start2 = System.currentTimeMillis()
-
+      val start = System.currentTimeMillis()
```
ellykits, line 117: You can use Kotlin's lambda function [measureTimeMillis](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.system/measure-time-millis.html)

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
```diff
@@ -150,30 +149,27 @@ constructor(
...
-                }
-              }
-              .toSet()
-          )
-          .also { ruleConfigsCache[ruleConfigsKey] = it }
-      )
-
-    return jexlRules
+  suspend fun generateRules(ruleConfigs: List<RuleConfig>): Rules {
+    val start = System.currentTimeMillis()
```
ellykits, line 153: You can use Kotlin's lambda function [measureTimeMillis](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.system/measure-time-millis.html)

ndegwamartin:
The user tests for this PR didn’t seem to produce any significant performance improvement for the app. The areas where we would have gained via concurrency the most were in DB interaction however sqlite DB on Android does not support multithreading in a straight forward way. There are some PRAGMA flags that could alter the way it works for us e.g. enabling [Write Ahead Logging](https://sqlite.org/wal.html) but there’s the trade off in that we risk durability i.e. loosing some data incase for instance a battery is removed from the device. On Android reading and writing from the db use same same connection under the hood when running the [recommended set up](https://developer.android.com/training/data-storage/sqlite). It may be possible to break the existing recommended pattern on Android to use distinct connections for reading and for writing from the db to gain performance which would involve some refactor for support on the SDK

**Links :** 
- [SQLite WAL](https://sqlite.org/wal.html)
- [SQLite Android dev](https://developer.android.com/training/data-storage/sqlite)
- [Android SQLite Locking](https://touchlabblog.tumblr.com/post/24474398246/android-sqlite-locking)

pld:
Makes sense

PR #2135, opened 2023-03-10
Fix StructureMap missing app crash

### LINKED ISSUE #1848: [Quest] Provide error-handling if StructureMap has an error during the extraction process

SebaMutuku:

Fixes #1848 

**Checklist**


android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/ConfigurationRegistry.kt
```diff
@@ -322,6 +322,7 @@ constructor(
         composition
           .retrieveCompositionSections()
           .groupBy { it.focus.reference?.split(TYPE_REFERENCE_DELIMITER)?.firstOrNull() ?: "" }
+          .asSequence()
```
ellykits, line 325: This can be moved before groupBy

android/engine/src/main/java/org/smartregister/fhircore/engine/domain/model/ActionConfig.kt
```diff
@@ -41,7 +41,7 @@ data class ActionConfig(
 ) {
   fun paramsBundle(computedValuesMap: Map<String, Any> = emptyMap()): Bundle =
     Bundle().apply {
-      params.filter { !it.paramType?.name.equals(PREPOPULATE_PARAM_TYPE) }.forEach {
+      params.asSequence().filter { !it.paramType?.name.equals(PREPOPULATE_PARAM_TYPE) }.forEach {
```
ellykits, line 44: I do not think this is really necessary here, there is only one operation.

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
```diff
@@ -375,9 +380,10 @@ constructor(
       if (fhirPathExpression.isEmpty()) {
         return emptyList()
       }
-      return resources?.filter {
-        fhirPathDataExtractor.extractValue(it, fhirPathExpression).toBoolean()
-      }
+      return resources
+        ?.asSequence()
```
ellykits, line 384: I don't think a sequence will add any performance improvement here there is only one filter operation. Collecting result for long Lists is not performant on a sequence.

android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirCarePlanGenerator.kt
```diff
@@ -192,8 +192,10 @@ constructor(
       if (carePlan.status == CarePlan.CarePlanStatus.COMPLETED)
         carePlan
           .activity
+          .asSequence()
           .flatMap { it.outcomeReference }
           .filter { it.reference.startsWith(ResourceType.Task.name) }
+          .toList()
```
ellykits, line 198: Why are we introducing this lambda here? The return type for a filter operation is already a list. This `toList` is redundant and will impact the performance of the operations

android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirTaskExpireUtil.kt
```diff
@@ -70,7 +70,9 @@ constructor(@ApplicationContext val appContext: Context, val fhirEngine: FhirEng
           count = tasksCount
           sort(Task.AUTHORED_ON, Order.ASCENDING)
         }
+        .asSequence()
         .filter { it.isPastExpiry() }
+        .toList()
```
ellykits, line 75: Here we have a stateful operation happening on each item of a list, so using a sequence here is not ideal. So converting this to a list is not required. See [Working with stateless/stateful operations](https://typealias.com/guides/when-to-use-sequences/)

android/engine/src/main/java/org/smartregister/fhircore/engine/util/extension/MeasureExtensions.kt
```diff
@@ -117,5 +117,7 @@ suspend inline fun <reified R : Resource> retrievePreviouslyGeneratedMeasureRepo
       filter(MeasureReport.MEASURE, { value = measureUrl })
       operation = queryOperation
     }
+    .asSequence()
     .filter { it.period.start.formatDate(SDF_YYYY_MM_DD) == startDateFormatted }
+    .toList()
```
ellykits, line 122: Generally I do not think we need to use a sequence when we eventually need to convert something back to a collection so this usage is not adding any performance improvement.

android/engine/src/test/java/org/smartregister/fhircore/engine/sync/SyncBroadcasterTest.kt
```diff
@@ -136,16 +136,20 @@ class SyncBroadcasterTest : RobolectricTest() {
...
-      Assert.assertTrue(syncParam[it]!!.containsKey("_count"))
-    }
+    syncParam
+      .keys
+      .asSequence()
+      .filter { it.isIn(ResourceType.Binary, ResourceType.StructureMap) }
+      .forEach { Assert.assertTrue(syncParam[it]!!.containsKey("_count")) }
 
-    syncParam.keys.filter { it.isIn(ResourceType.Patient) }.forEach {
+    syncParam.keys.asSequence().filter { it.isIn(ResourceType.Patient) }.forEach {
```
ellykits, line 145: Using as sequence here is not useful. You need to check the kind of operations that are used after the sequence. In this case we filter then run for each statement so no impact at all.

android/geowidget/src/main/java/org/smartregister/fhircore/geowidget/screens/GeoWidgetViewModel.kt
```diff
@@ -73,8 +73,10 @@ constructor(val defaultRepository: DefaultRepository, val dispatcherProvider: Di
       }
 
     val familiesWithLocations =
-      defaultRepository.fhirEngine
+      defaultRepository
+        .fhirEngine
         .search<Group> { filter(Group.TYPE, { value = of(coding) }) }
+        .asSequence()
```
ellykits, line 79: Sequence not useful here for single operation

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/main/AppMainViewModel.kt
```diff
@@ -118,7 +118,9 @@ constructor(
   }
 
   fun retrieveIconsAsBitmap() {
-    navigationConfiguration.clientRegisters
+    navigationConfiguration
+      .clientRegisters
+      .asSequence()
```
ellykits, line 123: sequence not useful for simple operation

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/main/AppMainViewModel.kt
```diff
@@ -239,7 +241,7 @@ constructor(
     countsMap: SnapshotStateMap<String, Long>
   ) {
     // Set count for registerId against its value. Use action Id; otherwise default to menu id
-    this.filter { it.showCount }.forEach { menuConfig ->
+    this.asSequence().filter { it.showCount }.forEach { menuConfig ->
```
ellykits, line 244: Sequence not useful for simple operation like `filter` and `forEach`

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/profile/ProfileViewModel.kt
```diff
@@ -262,6 +263,7 @@ constructor(
               )
               .toBoolean()
           }
+          ?.toList()
```
ellykits, line 266: Is this toList() necessary for the map operation below? if that is the case then I do not think it is of best interest to use a sequence here.

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireActivity.kt
```diff
@@ -101,11 +101,14 @@ open class QuestionnaireActivity : BaseMultiLanguageActivity(), View.OnClickList
...
-          !it.value.contains(STRING_INTERPOLATION_PREFIX)
-      }
+      actionParams
+        .asSequence()
+        .filter {
+          it.paramType == ActionParameterType.PREPOPULATE &&
+            !it.value.isNullOrEmpty() &&
+            !it.value.contains(STRING_INTERPOLATION_PREFIX)
+        }
+        .toList()
```
ellykits, line 111: Is this `toList` necessary? We only have a single filter operation in this sequence.

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireViewModel.kt
```diff
@@ -177,7 +177,7 @@ constructor(
     }
 
     viewModelScope.launch(dispatcherProvider.io()) {
-      questionnaire.useContext.filter { it.hasValueCodeableConcept() }.forEach {
+      questionnaire.useContext.asSequence().filter { it.hasValueCodeableConcept() }.forEach {
```
ellykits, line 180: I do no think there will be any performance improvements using a sequence here.

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/MeasureReportViewModel.kt
```diff
@@ -538,7 +539,9 @@ constructor(
         // the observations would have key as coding.code=populationId and value as codeableConcept
         observations
           .groupBy { it.codingOf(POPULATION_OBS_URL)?.display }
+          .asSequence()
           .filter { it.key.isNullOrBlank().not() }
+          .toList()
```
ellykits, line 544: Is this toList() necessary? Sequence is not performant when you collect the result

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/main/components/BottomScreenSection.kt
```diff
@@ -51,7 +51,7 @@ fun BottomScreenSection(
   ) {
     val navBackStackEntry by navController.currentBackStackEntryAsState()
     val currentRoute = navBackStackEntry?.destination?.id
-    mainNavigationScreens.filter { it.showInBottomNav }.forEach { navigationScreen ->
+    mainNavigationScreens.asSequence().filter { it.showInBottomNav }.forEach { navigationScreen ->
```
ellykits, line 54: I do not think there will be any performance improvement using a sequence here for just the `filter` and `forEach` operations.

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
```diff
@@ -214,10 +214,15 @@ constructor(
             facts.getFact(relatedResourceKey).value as List<Resource>
           else emptyList()
 
-      return value.filter {
-        resource.logicalId ==
-          fhirPathDataExtractor.extractValue(it, referenceFhirPathExpression).extractLogicalIdUuid()
-      }
+      return value
+        .asSequence()
```
ellykits, line 218: Using a sequence here is not useful when we eventually collect the results in a list.

ellykits (changes requested):

ellykits:
Kindly review this [Working with sequences](https://typealias.com/guides/when-to-use-sequences/) especially the section that addresses stateless and stateful operations.

pld:
The goal of using asSequence here and other places, and I'm not sure it's the correct tool, is to prevent multiple passes over the array, explicitly having a conditional in a loop for the filter would work too, not sure which is best

ellykits:
I understand the usage. There are scenarios where using a sequence improves performance however not in all cases.

android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/ConfigurationRegistry.kt
SebaMutuku: @ellykits Adding `asSequence()` before `groupBy` is redudant. You have a look.

android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirCarePlanGenerator.kt
SebaMutuku: @ellykits we can have a discussion about this.

android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirTaskExpireUtil.kt
SebaMutuku: Maybe we can discuss more about this on a call or huddle.
Here's a sample 
``` kotlin  
val result = collectionOf12Items
    .asSequence()
    .map { it.toString() }
    .toList() 
```

pld:
> I understand the usage. There are scenarios where using a sequence improves performance however not in all cases.

cool sequence isn't what I see as the important part, single pass is the important part, but maybe there's a pre-compilation or some other reason single pass isn't that important, I haven't seen literature on that

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireActivity.kt
SebaMutuku: @ellykits it might be needed since `prePopulationParams` is of type `List<ActionParamater>`. However, using a sequence converts the collection to a `Sequence<ActionParameter>`. Maybe we can ignore the asSequence() or convert it toList() before reading elements or traversing within the collection

ellykits (approved):

PR #2145, opened 2023-03-11
Fix Cyclic Interpolation Crash 🐛

ndegwamartin:
Fixes crashes when exceptions occur during interpolation of the app config files

**Checklist**


android/engine/src/main/java/org/smartregister/fhircore/engine/util/extension/StringExtensions.kt
```diff
@@ -42,15 +43,19 @@ fun String.interpolate(
...
+  try {
+    StringSubstitutor.replace(
+      this.replace(Pattern.quote(prefix).plus(".*?").plus(Pattern.quote(suffix)).toRegex()) {
+        it.value.replace("\\s+".toRegex(), "")
+      },
+      lookupMap,
+      prefix,
+      suffix
+    )
+  } catch (e: Exception) {
```
pld, line 55: Is there anything more specific to catch? Can we test an input that leads to this exception?

ellykits: From the docs the function throws `IllegalArgumentException` [StringSubtitutor#replace](https://commons.apache.org/proper/commons-text/apidocs/org/apache/commons/text/StringSubstitutor.html#replace-java.lang.Object-java.util.Map-java.lang.String-java.lang.String-)

ndegwamartin: We could narrow it down I think to respective Exceptions but basically any exception thrown here will cause the app to crash

pld: I'm not that familiar with the structure of Android apps, but Isn't that (if an exception is thrown the app will crash) true of a lot of the code in a lot of places?

Is there some higher up main loop where we can can catch all Exceptions? So the app usually returns an error message to the user when there's an unrecoverable error, instead of crashing.

And then here we catch exceptions that we expect to be thrown here.

ndegwamartin: Yeah you're correct in that it will crash unless caught higher up. That is unless we catch the exception in the function it will have to be done at each point of invocation. 

For a global all catch solution doing some research on that

pld: moved to this issue, https://github.com/opensrp/fhircore/issues/2156

android/engine/src/test/java/org/smartregister/fhircore/engine/util/extension/StringExtensionTest.kt
```diff
@@ -38,4 +38,23 @@ class StringExtensionTest {
...
+      templateString.interpolate(lookupMap)
+    )
+  }
+
+  @Test
+  fun stringInterpolateShouldThrowIllegalArgumentExceptionIfPrefixNull() {
+    val templateString = "{ \"saveFamilyButtonText\" : @{ family.button.save }, @{ missing } }"
+    val lookupMap = mapOf<String, Any>("family.button.save" to "save")
+
+    Assert.assertEquals(templateString, templateString.interpolate(lookupMap))
```
pld, line 58: this fails, because it's not generating an IllegalArgumentException, or any Exception. If it was it would return the unchanged string and the test would pass, but it's returning the changed string and the test is failing.

I'm not sure what the scenario is where we will generate the exception. Because the StringSubstitor has the flag the throw an IllegalArgumentException when it cannot find a string set to false.

What's the bug situation we're trying to account for with the change to `interpolate`?

ndegwamartin: The exception we should be checking for is `IllegalStateException`. Additionally the exception occurs for an interesting edge case caused by the content of the key-value lookup map data.

I've updated the exception handling and unit test case to align with the above

pld: awesome thanks, but this test is not testing throwing the exception, if you comment out the try the throw doesn't happen. The idea with the map on line 56 in the test was that if exception gets thrown the interpolation should not happen. Can you update the test so that the exception throw is triggered and the result is tested?

pld (changes requested):

android/engine/src/test/java/org/smartregister/fhircore/engine/util/extension/StringExtensionTest.kt
ndegwamartin: If we want to test the throwing of the exception directly then we need to remove the try catch.

Right now it is being tested implicitly because but I suppose there's no way of telling that from the test case data

ndegwamartin: If we want to test the throwing of the exception directly then we need to remove the try catch.

Right now it is being tested implicitly but I suppose there's no way of telling that from the test case data

ndegwamartin: I've updated the test case data to make it more apparent

pld: oh yea this is doing that, I consider this explicit :D looks great

pld (approved):

pld (approved):

pld (approved):

PR #2161, opened 2023-03-15
Catch All Exceptions

### LINKED ISSUE #2156: Catch all exceptions somewhere high up on the app loop
https://github.com/opensrp/fhircore/pull/2145#discussion_r1133417655

**Acceptance criteria**
- An unexpected exception gets caught, e.g. try removing the catch here, and check that this check catches that lower down exception, https://github.com/opensrp/fhircore/pull/2145

**Area path**
A list of ordered steps in the app on usage of the feature to support anyone testing it e.g. Code reviewer, QA 

The code reviewer can test this by (Non QA technical test)
1. Throwing a runtime exception from any part of the app e.g. on clicking Sync button 
2. Build the app and run
3. Login and trigger the Runtime exception

1. Add uncaught exceptions handler in the Android Entry point Application class
2.  Catch and log exceptions via Timber
3. Show generic exception to the user using _Toast_. _(No other UI component available from Application Class)_

ndegwamartin:

 Closes #2156

**Engineer Checklist**

**Code Reviewer Checklist**


android/engine/src/main/java/org/smartregister/fhircore/engine/sync/SyncBroadcaster.kt
```diff
@@ -59,6 +59,9 @@ constructor(
 ) {
 
   fun runSync(syncSharedFlow: MutableSharedFlow<SyncJobStatus>) {
+
+    throw java.lang.RuntimeException("WE are cRushing on you")
```
pld, line 63: 🤣

ndegwamartin:
This update captures _all uncaught exceptions_ in the app. 

It displays a Toast message informing the user that an error occured. It then redirects to the launch page in the case that the exception was thrown on the main thread (_meaning app was in the foreground_). This will therefore prevent any app crashes. For this reason, we need some more exception handling logic for reporting the errors once we have a crash manager tool installed.

Once we add a crash reporting tool is selected like _Firebase Crashlytics_ or _Sentry_ we'll need to update the method there to log the errors for remote debugging.

There is a caveat for when a message is thrown too early in the app e.g. in the Sync Broadcaster during sync where on catching the exception the app fails to redirect properly and has to be killed/restarted manually. The approach around this would be to (try) catch all the appropriate errors in such sections. Thorough QA could help catch as many of these context specific exceptions as possible.

The above configuration is active only when the app is in _Release mode_ (thus disabled for debug mode). It is expected that any Demos for the FHIR Core app would be ran from a **Release APK**. It is good practice to let the app crash in _Debug mode_ so that there is instant feedback for the engineering team to fix the issues.

If we decide to merge the PR as is, the above must be factored in else we'd need to define a separate workflow in the case of uncaught exceptions.

ndegwamartin:
Build depends on Tech Debt fix in PR https://github.com/opensrp/fhircore/pull/2179

android/engine/src/main/java/org/smartregister/fhircore/engine/sync/SyncBroadcaster.kt
```diff
@@ -59,6 +59,7 @@ constructor(
 ) {
 
   fun runSync(syncSharedFlow: MutableSharedFlow<SyncJobStatus>) {
+
```
pld, line 62:
```suggestion
```

pld (dismissed):
lgtm! can you please add tests

dubdabasoduba:
> This update captures _all uncaught exceptions_ in the app.
> 
> It displays a Toast message informing the user that an error occurred. It then redirects to the launch page in the case that the exception was thrown on the main thread (_meaning the app was in the foreground_). This will therefore prevent any app crashes. For this reason, we need some more exception-handling logic for reporting the errors once we have a crash manager tool installed.
> 
> Once we add a crash reporting tool that is selected like _Firebase Crashlytics_ or _Sentry_ we'll need to update the method there to log the errors for remote debugging.
> 
> There is a caveat for when a message is thrown too early in the app e.g. in the Sync Broadcaster during sync where on catching the exception the app fails to redirect properly and has to be killed/restarted manually. The approach around this would be to (try) catch all the appropriate errors in such sections. Thorough QA could help catch as many of these context-specific exceptions as possible.
> 
> The above configuration is active only when the app is in _Release mode_ (thus disabled for debug mode). It is expected that any Demos for the FHIR Core app would be run from a **Release APK**. It is good practice to let the app crash in _Debug mode_ so that there is instant feedback for the engineering team to fix the issues.
> 
> If we decide to merge the PR as is, the above must be factored in else we'd need to define a separate workflow in the case of uncaught exceptions.

@ndegwamartin did we agree on the crash reporting tool? Can we have both? I know we have had issues trying to give clients access to the firebase console. We could use that internally but then use sentry (which is deployed as a monitoring tool for the servers) for clients. 

I think we already have samples for both integrations running for the OpenSRP apps.

android/quest/src/main/java/org/smartregister/fhircore/quest/QuestApplication.kt
```diff
@@ -75,4 +83,28 @@ class QuestApplication : Application(), DataCaptureConfig.Provider, Configuratio
...
+   *
+   * TODO add crash reporting when a crash reporting tool is selected e.g. Fabric Crashlytics or
+   * Sentry
+   */
+  private fun handleUncaughtException(e: Throwable) {
+    showToast(this.getString(R.string.error_occurred))
+    Timber.e(e)
+
+    if (Looper.myLooper() == Looper.getMainLooper()) {
+      val intent = Intent(applicationContext, AppSettingActivity::class.java)
```
dubdabasoduba, line 104: My assumption is this would then redirect the user to the PIN or username/password login page since the App Ids were already entered into the app. If yes I think that's fine but we can consult the Design team. 

I think it would be better to redirect to the home page assuming the crash is happening somewhere else other that the home page. We could redirect to the Login pages if the crashes happen on the home page. We could then redirect to the setting page if they happen on the login pages.

My reasoning for the above proposal is to still give the use a change to use the app if some pieces of it are usable. 

I realise it might be too much or over engineering of some sought though.

ndegwamartin:
> 



> > This update captures _all uncaught exceptions_ in the app.
> > It displays a Toast message informing the user that an error occurred. It then redirects to the launch page in the case that the exception was thrown on the main thread (_meaning the app was in the foreground_). This will therefore prevent any app crashes. For this reason, we need some more exception-handling logic for reporting the errors once we have a crash manager tool installed.
> > Once we add a crash reporting tool that is selected like _Firebase Crashlytics_ or _Sentry_ we'll need to update the method there to log the errors for remote debugging.
> > There is a caveat for when a message is thrown too early in the app e.g. in the Sync Broadcaster during sync where on catching the exception the app fails to redirect properly and has to be killed/restarted manually. The approach around this would be to (try) catch all the appropriate errors in such sections. Thorough QA could help catch as many of these context-specific exceptions as possible.
> > The above configuration is active only when the app is in _Release mode_ (thus disabled for debug mode). It is expected that any Demos for the FHIR Core app would be run from a **Release APK**. It is good practice to let the app crash in _Debug mode_ so that there is instant feedback for the engineering team to fix the issues.
> > If we decide to merge the PR as is, the above must be factored in else we'd need to define a separate workflow in the case of uncaught exceptions.
> 
> @ndegwamartin did we agree on the crash reporting tool? Can we have both? I know we have had issues trying to give clients access to the firebase console. We could use that internally but then use sentry (which is deployed as a monitoring tool for the servers) for clients.
> 
> I think we already have samples for both integrations running for the OpenSRP apps.

I think ideally we should only have one - and probably Sentry especially if it supports RBAC for external teams access. It might also play better with privacy in the context of health apps.

android/quest/src/main/java/org/smartregister/fhircore/quest/QuestApplication.kt
pld: agreed, we should consider options here, moving to this issue, https://github.com/opensrp/fhircore/issues/2193

PR #2232, opened 2023-04-05
Refactor register search and support all search API filter operations

### LINKED ISSUE #2209: Refactor register search to use SDK Search API to perform database query
Refactor register search to use SDK search DSL APIs instead of applying filter on the paginated items via code (very slow with a lot of data). 

**Additional context**
The current search implementation works by loading all the register data via the paging3 library, then applying the[ paging3 filter data transform function](https://developer.android.com/topic/libraries/architecture/paging/v3-transform#filter-data) to filter the data. 
_How the current filter works:_
1. A list of rule names is provided via the [RegisterConfiguration.searchBar.computedRules](https://github.com/opensrp/fhircore/blob/414d97d20310da9f8ad14ef0b00cba5975f2dcd3/android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/register/RegisterConfiguration.kt#L34) 
2. For each of the item above extract the value from the computed values map of every register list item and compare it against the search term
3. Return a filtered list of ResourceData matching the condition above
4. Cache all the retrieved resource data for future use when performing search (subsequent search will be faster at the expense of more memory usage via the cache). Cache is cleared when the user exists the register.

PS: When searching register data, the result is not paginated but uses an infinite scroll which is not consistent with the figma design that uses pagination.

**Acceptance criteria**
Register search should return paginated search response when a user searches the register data. The search data queries should also be configurable.

**Area path**
1. Load household register for instance (or any register) 
2. Enter name of household to search in the search input box

- Provide search data queries via RegisterConfiguration
- On search term change, query the database using the above query configs
- Refactor `loadRegisterData` to query the database using the configured queries and finally paginate and display the result

Query? Do we want to support code level filters for search parameters that are not supported via the Search API? is it possible to make non-searchable params searchable via the SDK? e.g. if `Group.name` is not searchable can the `name` property be indexed to allow search?

ellykits:

Fixes #2209  or Closes #2209 

**Engineer Checklist**

**Code Reviewer Checklist**


android/engine/src/main/java/org/smartregister/fhircore/engine/util/extension/QuestionnaireExtension.kt
```diff
@@ -181,5 +181,6 @@ fun String.castToType(type: DataType): Type? {
     DataType.URI -> UriType(this)
     DataType.CODING -> this.tryDecodeJson<Coding>()
     DataType.QUANTITY -> this.tryDecodeJson<Quantity>()
+    else -> null /*TODO cast the (several) remaining Enumeration.DataTypes*/
```
pld, line 184: is null OK or should we throw here?

android/engine/src/main/java/org/smartregister/fhircore/engine/util/extension/SearchExtension.kt
```diff
@@ -47,81 +48,76 @@ fun Search.filterByResourceTypeId(
...
 
 fun Search.filterBy(dataQuery: DataQuery) {
-  when (dataQuery.filterType) {
-    Enumerations.SearchParamType.TOKEN -> filterToken(dataQuery)
-    Enumerations.SearchParamType.STRING -> filterString(dataQuery)
-    Enumerations.SearchParamType.DATE -> filterDate(dataQuery)
-    else ->
-      throw UnsupportedOperationException("Can not apply ${dataQuery.filterType} as search filter")
+  dataQuery.filterCriteria.forEach { filterCriterion: FilterCriterionConfig ->
+    when (filterCriterion) {
```
pld, line 52: do the conditions cover the possible types of `filterCriterion`? if not, is it better to ignore the filter (as these changes do) or throw?

pld (commented):
looks great!

android/engine/src/main/java/org/smartregister/fhircore/engine/util/extension/SearchExtension.kt
ellykits: The conditions cover all the possible types.

android/engine/src/main/java/org/smartregister/fhircore/engine/util/extension/QuestionnaireExtension.kt
ellykits: We'd update the implementation here (maybe during CHT) to handle rest of the relevant DataTypes but for now returning `null` is okay.

pld (dismissed):

pld (approved):

PR #2235, opened 2023-04-06
Custom search param for active groups

Rkareko:

Fixes #[issue number] or Closes #[issue number]

**Engineer Checklist**

**Code Reviewer Checklist**

### DIFF
--- a/android/engine/src/main/java/org/smartregister/fhircore/engine/di/FhirEngineModule.kt
+++ b/android/engine/src/main/java/org/smartregister/fhircore/engine/di/FhirEngineModule.kt
@@ -84,8 +84,8 @@ class FhirEngineModule {
       SearchParameter().apply {
         url = "http://example.com/SearchParameter/group-active"
         addBase("Group")
-        name = "activeGroup"
-        code = "activeGroup"
+        name = GROUP_ACTIVE_SEARCH_PARAM
+        code = GROUP_ACTIVE_SEARCH_PARAM
         type = Enumerations.SearchParamType.TOKEN
         expression = "Group.active"
         description = "Search the active field"
@@ -96,5 +96,6 @@ class FhirEngineModule {
 
   companion object {
     private const val QUEST_OKHTTP_CLIENT_TAG = "QuestOkHttpClient"
+    const val GROUP_ACTIVE_SEARCH_PARAM = "activeGroup"
   }
 }
--- a/android/engine/src/main/java/org/smartregister/fhircore/engine/di/FhirEngineModule.kt
+++ b/android/engine/src/main/java/org/smartregister/fhircore/engine/di/FhirEngineModule.kt
@@ -30,6 +30,8 @@ import dagger.hilt.InstallIn
 import dagger.hilt.android.qualifiers.ApplicationContext
 import dagger.hilt.components.SingletonComponent
 import javax.inject.Singleton
+import org.hl7.fhir.r4.model.Enumerations
+import org.hl7.fhir.r4.model.SearchParameter
 import org.smartregister.fhircore.engine.BuildConfig
 import org.smartregister.fhircore.engine.configuration.app.ConfigService
 import org.smartregister.fhircore.engine.data.remote.shared.TokenAuthenticator
@@ -68,11 +70,28 @@ class FhirEngineModule {
                 headersToIgnore = listOf(AUTHORIZATION, COOKIE)
               )
             ) { Timber.tag(QUEST_OKHTTP_CLIENT_TAG).d(it) }
-        )
+        ),
+          customSearchParameters = getCustomSearchParameters()
+
       )
     )
-
+    Timber.e("Fhirengine module custom params set +++++++++++++++++++")
     return FhirEngineProvider.getInstance(context)
+  }
+
+  fun getCustomSearchParameters(): List<SearchParameter> {
+    val activeGroupSearchParameter =
+      SearchParameter().apply {
+        url = "http://example.com/SearchParameter/group-active"
+        addBase("Group")
+        name = "activeGroup"
+        code = "activeGroup"
+        type = Enumerations.SearchParamType.TOKEN
+        expression = "Group.active"
+        description = "Search the active field"
+      }
+
+    return listOf(activeGroupSearchParameter)
   }
 
   companion object {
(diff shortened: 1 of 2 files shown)


android/engine/src/main/java/org/smartregister/fhircore/engine/di/FhirEngineModule.kt
```diff
@@ -68,14 +70,31 @@ class FhirEngineModule {
...
+        expression = "Group.active"
+        description = "Search the active field"
+      }
+
+    return listOf(activeGroupSearchParameter)
+  }
+
   companion object {
     private const val QUEST_OKHTTP_CLIENT_TAG = "QuestOkHttpClient"
+    const val GROUP_ACTIVE_SEARCH_PARAM = "activeGroup"
```
ellykits, line 98: For consistency I would suggest we call this `active` as the name of the property in the [FHIR Spec Group resource](https://build.fhir.org/group.html)

```suggestion
    const val ACTIVE_SEARCH_PARAM = "active"
```

android/engine/src/main/java/org/smartregister/fhircore/engine/di/FhirEngineModule.kt
```diff
@@ -68,14 +70,31 @@ class FhirEngineModule {
                 headersToIgnore = listOf(AUTHORIZATION, COOKIE)
               )
             ) { Timber.tag(QUEST_OKHTTP_CLIENT_TAG).d(it) }
-        )
+        ),
+        customSearchParameters = getCustomSearchParameters()
       )
     )
-
+    Timber.e("Fhirengine module custom params set +++++++++++++++++++")
```
ellykits, line 77: Delete this log

ellykits (dismissed):

android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/app/ConfigService.kt
```diff
@@ -66,4 +68,24 @@ interface ConfigService {
...
+        type = Enumerations.SearchParamType.TOKEN
+        expression = "Group.active"
+        description = "Search the active field"
+      }
+
+    return listOf(activeGroupSearchParameter)
+  }
+
+  companion object {
+    const val GROUP_ACTIVE_SEARCH_PARAM = "activeGroup"
```
pld, line 89: per @ellykits' suggestion 
```suggestion
    const val GROUP_ACTIVE_SEARCH_PARAM = "active"
```

pld:
this also looks like it needs explicit testing for the resource type group being filtered and not, interesting that the new lines are showing that they're covered, what's missing is at least some test data with active and inactive groups and asserts verify only the active groups remain after filtering

android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/app/ConfigService.kt
Rkareko: updated

android/engine/src/main/java/org/smartregister/fhircore/engine/di/FhirEngineModule.kt
```diff
@@ -68,14 +70,31 @@ class FhirEngineModule {
...
+  fun getCustomSearchParameters(): List<SearchParameter> {
+    val activeGroupSearchParameter =
+      SearchParameter().apply {
+        url = "http://example.com/SearchParameter/group-active"
+        addBase("Group")
+        name = GROUP_ACTIVE_SEARCH_PARAM
+        code = GROUP_ACTIVE_SEARCH_PARAM
+        type = Enumerations.SearchParamType.TOKEN
+        expression = "Group.active"
+        description = "Search the active field"
```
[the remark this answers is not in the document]
Rkareko: updated

dubdabasoduba (dismissed):

android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/app/ConfigService.kt
```diff
@@ -66,4 +68,24 @@ interface ConfigService {
   }
 
   fun provideConfigurationSyncPageSize(): String
+
+  /** Provide a list of custom search parameters */
+  fun provideCustomSearchParameters(): List<SearchParameter> {
+    val activeGroupSearchParameter =
+      SearchParameter().apply {
+        url = "http://smartregister.org/SearchParameter/group-active"
```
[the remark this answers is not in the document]
Rkareko: @dubdabasoduba yes it's mandatory according to the [Search Parameter documentation](https://build.fhir.org/searchparameter-definitions.html#SearchParameter.url)

Any suggestions are welcome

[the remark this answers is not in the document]
pld: my understanding is that URL here is like the URL for a code, where it's referring to an endpoint that defines a spec, in which it's ok for it to be constant b/c the spec should live at a constant place -- of course we don't actually have a spec at this URL, but like with codes

pld (changes requested):
other than missing tests, lgtm

android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/app/ConfigService.kt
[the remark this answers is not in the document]
dubdabasoduba: @pld Yes, that's true. I guess my concern is making it hard-coded. It means a client can't change it easily if they have their own spec definitions. 

We can clean it up later though.

pld:
Yea I think we can open an issue for that, because I feel like to make that worthwhile we’d need to load some functionality from the URL in addition to letting a client configure it.

pld (approved):
perfect

android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/app/ConfigService.kt
```diff
@@ -66,4 +68,24 @@ interface ConfigService {
   }
 
   fun provideConfigurationSyncPageSize(): String
+
+  /** Provide a list of custom search parameters */
```
pld, line 72:
```suggestion
  /** 
   * Provide a list of custom search parameters.
   *
   * @return list of predefined custom group search parameters.
   */
```

pld (approved):

pld (approved):

Issue #2252, opened 2023-04-18
[Enhancement] Change button's background color property


Lentumunai-Mark:
**Describe the feature request.**
One may want to be able to change the background color property for a button, by passing the background color attribute in the service-card properties in the config. Currently, one can specify the background color however, it is not reflected in the button.

**Additional context**
This feature will be helpful in scenarios where the user wants to customize the background color.

**Acceptance criteria**
If one specifies the background color for the button then the button should be able to reflect the given color.

**Area path**
Open the config files and add the new property background color to the service card property where applicable.
Open the app and navigate to the page where the config above is applicable.
Verify that the background color behavior is consistent with the color you applied.

**Implementation plan (For Engineers)**
Interpolate the background color.
Apply the color above to the button's background.

PR #2253, opened 2023-04-18
Change button background color enhancement

Lentumunai-Mark:

Fixes https://github.com/opensrp/fhircore/issues/2252

**Engineer Checklist**

**Code Reviewer Checklist**


android/quest/src/main/java/org/smartregister/fhircore/quest/ui/shared/components/ActionableButton.kt
```diff
@@ -178,6 +183,14 @@ fun ButtonProperties.interpolateStatus(computedValuesMap: Map<String, Any>): Ser
   else ServiceStatus.UPCOMING
 }
 
+@Composable
+fun ButtonProperties.interpolateBackgroundColor(computedValuesMap: Map<String, Any>): Color {
+  val interpolated = this.backgroundColor?.interpolate(computedValuesMap)
+  return if (!interpolated.isNullOrEmpty()) {
+    return interpolated.parseColor()
+  } else Color.Unspecified
+}
```
pld, line 192: what if combine interpolation for any (or all the ones we want to) ButtonProperties into a single step, any thoughts on how we can do that?

relatedly, wondering if these interpolation functions (or a more general one) should live in ButtonProperties and not here.

relatedly, a little uneasy that we're combining interpolation and setting a default (with the else) into single step, that's a bit broader than only interpolation b/c the name is mislead -- I see this pattern was introduced before so this is not a critique of this specific addition, it's following that pattern

Lentumunai-Mark: Yeah sure, I'll move the interpolations to one function in the button properties responsible for handling all the interpolations.

pld (changes requested):
lgtm!

pld (approved):

roywanyaga (approved):
looks good to me

PR #2257, opened 2023-04-18
Render profile view incrementally

### LINKED ISSUE #2165: Render profile view in stages
**Describe the Issue**
Render the profile view in stages so that the page is not left blank for a period of time. The page will be rendered as soon as it receives data.

- Split loading and rendering of data so that the demographic data is shown immediately it's rendered. 
- Split the rendering of list items in the profile page so that each list item is fetched, run through the rules engine to calculate the values to be shown and then rendered immediately before moving to the item. This assumes that each list items has related resources that take time to load thus the difference in UX when rendered individually given that the primary list of resources is loaded from a single database query

ekigamba:

Fixes #2165

**Engineer Checklist**

**Code Reviewer Checklist**


android/engine/src/main/java/org/smartregister/fhircore/engine/data/local/register/RegisterRepository.kt
```diff
@@ -216,6 +220,93 @@ constructor(
     return relatedResourcesData
   }
 
+  private suspend fun searchRelatedResourcesAndProcess(
```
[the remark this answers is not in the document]
pld: what's the function it's duplicating?

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/shared/components/ViewRenderer.kt
```diff
@@ -49,6 +51,7 @@ fun ViewRenderer(
   resourceData: ResourceData,
   navController: NavController
 ) {
+  Timber.e("Regenerating the views")
```
pld, line 54: remove before merge?

ekigamba: Yes

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/main/AppMainActivity.kt
```diff
@@ -116,7 +116,8 @@ open class AppMainActivity : BaseMultiLanguageActivity(), QuestionnaireHandler,
     // Setup the drawer and schedule jobs
     appMainViewModel.run {
       retrieveAppMainUiState()
-      schedulePeriodicJobs()
+      // schedulePeriodicJobs()
+      workManager.cancelAllWork()
```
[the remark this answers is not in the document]
ekigamba: Definitely.

android/engine/src/main/java/org/smartregister/fhircore/engine/data/local/register/RegisterRepository.kt
```diff
@@ -190,7 +193,7 @@ constructor(
       fhirPathDataExtractor
         .extractData(baseResource, fhirPathExpression)
         .takeWhile { it is Reference }
-        .pmap {
+        .map {
```
ellykits, line 196: Any reason for not using `pmap` here?

android/engine/src/main/java/org/smartregister/fhircore/engine/data/local/register/RegisterRepository.kt
```diff
@@ -216,6 +219,95 @@ constructor(
     return relatedResourcesData
   }
 
+  private suspend fun searchRelatedResourcesAndProcess(
+    resourceConfig: ResourceConfig,
+    baseResourceType: ResourceType,
+    baseResource: Resource,
+    fhirPathExpression: String?,
+    queryResult: RepositoryResourceData.QueryResult.Search,
+    afterFetch: (suspend () -> Unit)
```
ellykits, line 228: I think there is possibility of leaking the ProfileViewModel where this `afterFetch` function is implemented. The function is called in a repository class which might out live the viewmodel if implemented as a singleton.

android/engine/src/main/java/org/smartregister/fhircore/engine/domain/model/ResourceData.kt
```diff
@@ -32,6 +39,8 @@ import org.hl7.fhir.r4.model.ResourceType
 data class ResourceData(
   val baseResourceId: String,
   val baseResourceType: ResourceType,
-  val computedValuesMap: Map<String, Any>,
-  val listResourceDataMap: Map<String, List<ResourceData>>? = null
+  val computedValuesMap: SnapshotStateMap<String, Any>,
+  val listResourceDataMap: SnapshotStateMap<String, SnapshotStateList<ResourceData>> =
+    mutableStateMapOf(),
+  val baseResource: Resource? = null
```
ellykits, line 45: ResourceData  here will hold reference to the base resource which is not used anywhere in the rendered view thus leaking memory at this point.

android/engine/src/main/java/org/smartregister/fhircore/engine/domain/repository/Repository.kt
```diff
@@ -45,15 +47,41 @@ interface Repository {
...
+   */
+  suspend fun loadProfileRelatedAndSecondaryResources(
+    queryResult: RepositoryResourceData.QueryResult.Search,
+    listResourceDataMapState: SnapshotStateMap<String, SnapshotStateList<ResourceData>>,
+    resourceData: ResourceData,
+    profileId: String,
+    resourceId: String,
+    fhirResourceConfig: FhirResourceConfig?,
+    paramsList: Array<ActionParameter>?,
+    afterFetch: suspend () -> Unit
```
ellykits, line 85: I also noticed this afterFetch function is called multiple times (as many as the number of items iterated in the LIST view). Internally it has a forEach function that searches for all the LIST properties of the profile view adding to the number of calls.

android/engine/src/main/java/org/smartregister/fhircore/engine/data/local/register/RegisterRepository.kt
ekigamba: We realized that it doesn't improve performance. We also need this to happen sequentially

android/engine/src/main/java/org/smartregister/fhircore/engine/data/local/register/RegisterRepository.kt
ekigamba: Nice catch

android/engine/src/main/java/org/smartregister/fhircore/engine/domain/model/ResourceData.kt
ekigamba: Probably, :grimacing: we are probably using more memory but not leaking it. We need also need to the baseResource for the profile rendering and recomputations

android/engine/src/main/java/org/smartregister/fhircore/engine/domain/repository/Repository.kt
ekigamba: Yes. It's a very dirty solution written with the assumption that n will not grow and the cost of repeating these operations is small enough compared to rewriting the calculations and resource loading to be accurate enough to recalculate only operations that affect the recently loaded resources

ellykits:
We can close this PR in favor of https://github.com/opensrp/fhircore/pull/2354. Most of the issues raised here are addressed on the mentioned PR

ellykits:
Used a[ SnapshotStateList<ResourceData> ](https://developer.android.com/reference/kotlin/androidx/compose/runtime/snapshots/SnapshotStateList)on the list view. Compose is able to refresh the UI when an item is added to the list.

PR #2262, opened 2023-04-19
Update Pin Login Screen Message 

sharon2719:

Fixes #[issue number] or Closes #[issue number]

**Engineer Checklist**

**Code Reviewer Checklist**

### DIFF
--- a/android/quest/src/test/java/org/smartregister/fhircore/quest/ui/pin/PinViewModelTest.kt
+++ b/android/quest/src/test/java/org/smartregister/fhircore/quest/ui/pin/PinViewModelTest.kt
@@ -62,6 +62,8 @@ class PinViewModelTest : RobolectricTest() {
     every { secureSharedPreference.retrieveSessionUsername() } returns "demo"
     pinViewModel.setPinUiState(true, context)
     val pinUiState = pinViewModel.pinUiState.value
+    val expectedMessage = "VHT will use this PIN to login"
+    Assert.assertNotNull(expectedMessage, pinUiState.message)
     Assert.assertEquals("1245", pinUiState.currentUserPin)
     Assert.assertTrue(pinUiState.setupPin)
   }
--- a/android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/app/LoginConfig.kt
+++ b/android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/app/LoginConfig.kt
@@ -22,5 +22,6 @@ import kotlinx.serialization.Serializable
 data class LoginConfig(
   val showLogo: Boolean = true,
   val enablePin: Boolean? = false,
-  val pinLength: Int = 4
+  val pinLength: Int = 4,
+  val pinLoginMessage: String? = null
 )
--- a/android/quest/src/main/java/org/smartregister/fhircore/quest/ui/pin/PinViewModel.kt
+++ b/android/quest/src/main/java/org/smartregister/fhircore/quest/ui/pin/PinViewModel.kt
@@ -85,7 +85,9 @@ constructor(
         currentUserPin = secureSharedPreference.retrieveSessionPin() ?: "",
         setupPin = setupPin,
         message =
-          if (setupPin) context.getString(R.string.set_pin_message)
+          if (setupPin)
+            applicationConfiguration.loginConfig.pinLoginMessage
+              ?: context.getString(R.string.set_pin_message)
           else context.getString(R.string.enter_pin_for_user, username),
         pinLength = applicationConfiguration.loginConfig.pinLength,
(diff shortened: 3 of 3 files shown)


android/quest/src/test/java/org/smartregister/fhircore/quest/ui/pin/PinViewModelTest.kt
```diff
@@ -62,6 +62,8 @@ class PinViewModelTest : RobolectricTest() {
     every { secureSharedPreference.retrieveSessionUsername() } returns "demo"
     pinViewModel.setPinUiState(true, context)
     val pinUiState = pinViewModel.pinUiState.value
+    val expectedMessage = "VHT will use this PIN to login"
+    Assert.assertNotNull(expectedMessage, pinUiState.message)
```
pld, line 66: shouldn't this be assertEquals?

sharon2719: @pld  I have revised it to AssertEquals. Users can now configure and set their PinLoginScreen message to that of their choice

android/quest/src/test/java/org/smartregister/fhircore/quest/ui/pin/PinViewModelTest.kt
```diff
@@ -58,14 +61,28 @@ class PinViewModelTest : RobolectricTest() {
...
+        )
+      )
     every { secureSharedPreference.retrieveSessionPin() } returns "1245"
     every { secureSharedPreference.retrieveSessionUsername() } returns "demo"
-    pinViewModel.setPinUiState(true, context)
-    val pinUiState = pinViewModel.pinUiState.value
-    Assert.assertEquals("1245", pinUiState.currentUserPin)
-    Assert.assertTrue(pinUiState.setupPin)
+    pinViewModel.pinUiState.value = pinLoginState.value
+    val expectedMessage = "VHT will use this PIN to login"
```
pld, line 79: to reduce duplication and guarantee equality, please move this `val` above the definition of PinUiState and then pass it as an argument on initialization

pld (changes requested):

pld (approved):

android/quest/src/test/java/org/smartregister/fhircore/quest/ui/pin/PinViewModelTest.kt
```diff
@@ -58,12 +61,25 @@ class PinViewModelTest : RobolectricTest() {
...
     every { secureSharedPreference.retrieveSessionPin() } returns "1245"
     every { secureSharedPreference.retrieveSessionUsername() } returns "demo"
-    pinViewModel.setPinUiState(true, context)
-    val pinUiState = pinViewModel.pinUiState.value
-    Assert.assertEquals("1245", pinUiState.currentUserPin)
-    Assert.assertTrue(pinUiState.setupPin)
+    pinViewModel.pinUiState.value = pinLoginState.value
+    Assert.assertEquals(expectedMessage, pinLoginState.value.message)
+    Assert.assertEquals("1245", pinLoginState.value.currentUserPin)
+    Assert.assertTrue(pinLoginState.value.setupPin)
```
pld, line 82: on closer inspection, this is not testing what we want it to. This is testing the value of pinLoginState, but we just set that above on line 66, so we know what the values will be. What we would like to test here are the values of the pinViewModel.

Relatedly, we need to make these tests cover the newly uncovered line. But, because PinUIState can not have a nullable message, we would have to set the pinLoginMessage in the app config to null and pass that through for processing, to verify that when it is null, we get the default value back.

Hope this helps, please let us know if you need more guidance on rewriting these tests.

pld (changes requested):

android/quest/src/test/java/org/smartregister/fhircore/quest/ui/pin/PinViewModelTest.kt
sharon2719: @pld I would require more guidance on rewriting this since its essential for the eCHIS project

ellykits: Schedule a meeting with me Sharon on Tuesday. I will assist you.

ndegwamartin: @sharon2719 also wherever can we need to follow the convention below for test case naming as documented in - https://fhircore.smartregister.org/developer-setup/code-testing

`<methodName><ConditionUnderTest><ExpectedBehavior>`

pld (approved):
tests look good!

PR #2289, opened 2023-04-26
Update Google FHIR SDK dependencies

allan-on:

Updated the Engine, SDC, Workflow and Contrib:Barcode dependencies to the latest released versions

See https://github.com/google/android-fhir/compare/master...opensrp:android-fhir:master-release

**Engineer Checklist**

**Code Reviewer Checklist**
- [x] I have built and run the FHIRCore app to verify the change fixes the issue and/or does not break the app

### DIFF
--- a/android/engine/src/main/java/org/smartregister/fhircore/engine/util/extension/FhirEngineExtension.kt
+++ b/android/engine/src/main/java/org/smartregister/fhircore/engine/util/extension/FhirEngineExtension.kt
@@ -54,11 +54,12 @@ suspend fun FhirEngine.loadLibraryAtPath(fhirOperator: FhirOperator, path: Strin
     runCatching { get<Library>(IdType(path).idPart) }.getOrNull()
       ?: search<Library> { filter(Library.URL, { value = path }) }.firstOrNull()
 
-  library?.let {
+  // TODO Fix refactored implementation on fhirOperator.loadLib()
+  /*library?.let {
     fhirOperator.loadLib(it)
 
     it.relatedArtifact.forEach { loadLibraryAtPath(fhirOperator, it) }
-  }
+  }*/
 }
 
 suspend fun FhirEngine.loadLibraryAtPath(
--- a/.github/workflows/sonarqube.yml
+++ b/.github/workflows/sonarqube.yml
@@ -1,0 +1,63 @@
+name: SonarQube Analysis
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    types: [opened, synchronize, reopened]
+    branches:
+      - main
+
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0  # Shallow clones should be disabled for a better relevancy of analysis
+
+      - name: Set up JDK 11
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+
+      - name: Cache SonarQube packages
+        uses: actions/cache@v1
+        with:
+          path: ~/.sonar/cache
+          key: ${{ runner.os }}-sonar
+          restore-keys: ${{ runner.os }}-sonar
+
+      - name: Cache Gradle packages
+        uses: actions/cache@v1
+        with:
+          path: ~/.gradle/caches
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
+          restore-keys: ${{ runner.os }}-gradle
+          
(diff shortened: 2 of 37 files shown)


android/engine/src/main/java/org/smartregister/fhircore/engine/util/extension/FhirEngineExtension.kt
```diff
@@ -54,11 +54,12 @@ suspend fun FhirEngine.loadLibraryAtPath(fhirOperator: FhirOperator, path: Strin
     runCatching { get<Library>(IdType(path).idPart) }.getOrNull()
       ?: search<Library> { filter(Library.URL, { value = path }) }.firstOrNull()
 
-  library?.let {
+  // TODO Fix refactored implementation on fhirOperator.loadLib()
+  /*library?.let {
     fhirOperator.loadLib(it)
 
     it.relatedArtifact.forEach { loadLibraryAtPath(fhirOperator, it) }
```
pld, line 61: can we remove these lines? if it's ok to comment them, why are they needed?

pld (changes requested):

android/engine/src/main/java/org/smartregister/fhircore/engine/util/extension/FhirEngineExtension.kt
ellykits: I believe that can be addressed separately

pld: OK, but if this means we will not be able to evaluate measures, we can't merge this without the fix

ellykits: Yea. @allan-on can you tests this? I will also tag @maimoonak.

ellykits: Yea. @allan-on can you tests this? I will also tag @maimoonak.

pld: K thanks, and if we still can evaluate measures, then I think we can remove these lines until we need to use `loadLib` and address them then

allan-on: Removing these lines will break evaluation of measures. 

In summary, the changes on the SDK introduce loading of Resources, for instance, a Library that is used for evaluation of Immunizations, from the [KnowledgeManager](knowledge/src/main/java/com/google/android/fhir/knowledge/KnowledgeManager.kt). The KnowledgeManager persists them in a separate database and when we do evaluate any Measure there's a call to  the `knowledgeManager.loadResources()`. 

So in this specific implementation, the `loadLib()` was removed and instead there's [install(implementationGuide: ImplementationGuide, rootDirectory: File)](https://github.com/google/android-fhir/blob/master/knowledge/src/main/java/com/google/android/fhir/knowledge/KnowledgeManager.kt#L58 )
or [install(file: File)](https://github.com/google/android-fhir/blob/master/knowledge/src/main/java/com/google/android/fhir/knowledge/KnowledgeManager.kt#L76)

The refactors required on our end would probably need to make use of the provided implementations. 

From my understanding the resources required will already be available in the DB but in a different `ResourceEntity` table. 

cc @ellykits @pld @maimoonak

pld: OK, that's reasonable, and means we need to resolve this TODO before we can merge this PR

allan-on: The other alternative to get the rest of the changes in faster would be to hold off on updating the `workflow` dependency version, and then draft an implementation plan for using the `KnowledgeManager` for Knowledge Resources essentially packaged up in an Implementation Guide such as ValueSets, Plan/Activity Definitions and Library Resources containing CQL.

pld:
That approach makes sense to me

f-odhiambo:
Passed QA for sync failure at 100%

android/engine/src/main/java/org/smartregister/fhircore/engine/util/extension/FhirEngineExtension.kt
pld: Let's do that so we can get the other updates merged, then have a separate PR to update workflow and integrate that w/measure eval

ndegwamartin (approved):

PR #2173, opened 2023-03-17
Fix QR not mapping answers

FikriMilano:
**IMPORTANT: Where possible all PRs must be linked to a Github issue**

Fixes #[issue number] or Closes #[issue number]

Changes:
- Fix QR not found in DB
- Remove initialExpression from a Questionnaire Item if that Questionnaire Item already has configurable pre-population values
- Fix FHIR Path initialExpressions for Family Questionnaire and Family Member Questionnaire
- Refactor ProfileViewModel and QuestionnaireActivity to error handling in questionnaire flow

**Engineer Checklist**
- [ ] I have written **Unit tests** for any new feature(s) and edge cases for bug fixes
- [ ] I have added any strings visible on UI components to the `strings.xml` file
- [ ] I have updated the  [CHANGELOG.md](./CHANGELOG.md) file for any notable changes to the codebase
- [ ] I have run `./gradlew spotlessApply` and `./gradlew spotlessCheck` to check my code follows the project's style guide
- [ ] I have built and run the FHIRCore app to verify my change fixes the issue and/or does not break the app 

**Code Reviewer Checklist**
- [ ] I have verified **Unit tests** have been written for any new feature(s) and edge cases
- [ ] I have verified any strings visible on UI components are in the `strings.xml` file
- [ ] I have verifed the [CHANGELOG.md](./CHANGELOG.md) file has any notable changes to the codebase
- [ ] I have verified the solution has been implemented in a configurable and generic way for reuseable components
- [ ] I have built and run the FHIRCore app to verify the change fixes the issue and/or does not break the app

PR #2327, opened 2023-05-09
Fix initial expression computation (Refactoring)

aurangzaibumer:
Initial changes can be found here : https://github.com/opensrp/fhircore/pull/2173#issue-1629292812

fixes https://github.com/opensrp/fhircore/issues/2335

[https://github.com/opensrp/fhircore/blob/96d8dffbe3b192927c4d45512beae8d55c98046f/[…]gister/fhircore/quest/ui/questionnaire/QuestionnaireActivity.kt](https://github.com/opensrp/fhircore/blob/96d8dffbe3b192927c4d45512beae8d55c98046f/android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireActivity.kt#L150)

_here's snippet just to check initial value was getting loaded_ 

**Engineer Checklist**

**Code Reviewer Checklist**


aurangzaibumer:
Just had a discussion with Fikri

Default is used when launching a Questionnaire to register something new.
Edit is used when launching a Questionnaire to edit something already exist.
Read Only is used when launching a Questionnaire in read only mode, means the user can’t edit

android/quest/src/main/assets/configs/app/profiles/default_profile_config.json
```diff
@@ -2235,6 +2235,7 @@
           "workflow": "LAUNCH_QUESTIONNAIRE",
           "questionnaire": {
             "id": "3276f55c-b25e-455b-ae4e-8846fb8fd039",
+            "type": "EDIT",
```
ekigamba, line 2238: I believe that this should be the fix for the conversation that we were having on Slack?

android/engine/src/main/java/org/smartregister/fhircore/engine/util/extension/QuestionnaireExtension.kt
```diff
@@ -146,16 +146,18 @@ fun List<Questionnaire.QuestionnaireItemComponent>.prePopulateInitialValues(
...
-    }
     prePopulationParams
       .firstOrNull {
         it.linkId == item.linkId &&
           !it.value.isNullOrEmpty() &&
           !it.value.contains(interpolationPrefix)
       }
       ?.let { actionParam ->
+
+        if (item.hasExtension(ITEM_INITIAL_EXPRESSION_URL)) {
```
ekigamba, line 157: We don't want to ever remove the initial expressions

ekigamba (changes requested):

android/engine/src/main/java/org/smartregister/fhircore/engine/util/extension/QuestionnaireExtension.kt
ekigamba: I believe it was being removed in favour of the prepopulation configs. Ideally, we should not have both the prepolutaion config for a specific question and have the same question with an initialExpression since they both do the same thing. I might be wrong however

pld (dismissed):

aurangzaibumer:
for some initial expression

Getting this exception
java.lang.NullPointerException: Attempt to invoke virtual method 'boolean org.hl7.fhir.r4.model.Base.isResource()' on a null object reference
                                                                                                    	at org.hl7.fhir.r4.utils.FHIRPathEngine.evaluate(FHIRPathEngine.java:525)
                                                                                                    	at com.google.android.fhir.datacapture.mapping.ResourceMapper.populateInitialValue(ResourceMapper.kt:246)
                                                                                                    	at com.google.android.fhir.datacapture.mapping.ResourceMapper.populateInitialValues(ResourceMapper.kt:232)
                                                                                                    	at com.google.android.fhir.datacapture.mapping.ResourceMapper.populate(ResourceMapper.kt:222)
                                                                                                    	at org.smartregister.fhircore.quest.ui.questionnaire.QuestionnaireViewModel.populateQuestionnaireResponse(QuestionnaireViewModel.kt:675)
                                                                                                    	at org.smartregister.fhircore.quest.ui.questionnaire.QuestionnaireViewModel.getQuestionnaireResponseFromDbOrPopulation(QuestionnaireViewModel.kt:658)
                                                                                                    	at org.smartregister.fhircore.quest.ui.questionnaire.QuestionnaireViewModel$getQuestionnaireResponseFromDbOrPopulation$1.invokeSuspend(Unknown Source:21)
                                                                                                    	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
                                                                                                    	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
                                                                                                    	at android.os.Handler.handleCallback(Handler.java:938)
                                                                                                    	at android.os.Handler.dispatchMessage(Handler.java:99)
                                                                                                    	at android.os.Looper.loopOnce(Looper.java:201)
                                                                                                    	at android.os.Looper.loop(Looper.java:288)
                                                                                                    	at android.app.ActivityThread.main(ActivityThread.java:7839)
                                                                                                    	at java.lang.reflect.Method.invoke(Native Method)
                                                                                                    	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
                                                                                                    	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1003)
                                                                                                    	Suppressed: kotlinx.coroutines.DiagnosticCoroutineContextException: [StandaloneCoroutine{Cancelling}@e54aeae, Dispatchers.Main.immediate]

pld:
yea I noticed that while looking through the test results too, @ekigamba any ideas?

ekigamba:
@aurangzaibumer This test `QuestionnaireExtensionTest > testFindQuestionnaireItemComponentPrepopulateRemovesInitialExpression` no longer serves it's purpose since we do not remove `intialExpression` even during prepopulation but I still need to understand the scenario under which the NPE is thrown since this is different I believe. 

You can report the results of your investigation when done

aurangzaibumer:
> @aurangzaibumer This test `QuestionnaireExtensionTest > testFindQuestionnaireItemComponentPrepopulateRemovesInitialExpression` no longer serves it's purpose since we do not remove `intialExpression` even during prepopulation but I still need to understand the scenario under which the NPE is thrown since this is different I believe.
> 
> You can report the results of your investigation when done

Yes - done my investigation so the NPE is coming from the ResourceMapper because the initialExpression needs Patient (when expression uses the patient resource) and it doesn't have that. More importantly the questionnaire that are being opened from the Visits list (in Patient profile) always come up with a subjectType `Task` - and if questionnaire has an initialExpression that tries to compute something through the `Patient` resource, it would not find that and hence crash appears

Here's in the example when we are trying to open one of the Task (through Visit List that is in the patient profile)
image

subjectType and subjectId are being sent through the bundle under the name/key of `baseResourceId` and `baseResourceType` - Now if we have an initialExpression in the questionnaire that uses `x` resource - we must have that resource object and its ID that needs in this method `getQuestionnaireResponseFromDbOrPopulation` so we can extract and send it the `ResourceMapper` class in order to be computed.

aurangzaibumer:
> 

Checked on older releases (0.1.9rc) - it used to be done this way to find the Array of Resources
image

pld (dismissed):

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireViewModel.kt
```diff
@@ -649,12 +650,28 @@ constructor(
...
+        if (questionnaireConfig.resourceIdentifier != null &&
+            questionnaireConfig.resourceType != null
+        ) {
+          populationResources =
+            loadPopulationResources(
+              questionnaireConfig.resourceIdentifier!!,
+              questionnaireConfig.resourceType!!
+            )
+        }
+      }
```
ndegwamartin, line 673: @aurangzaibumer looks like the previous implementation specifically tested against Tasks as a base resource being supported with the test case `testQRFromPopulationWithTaskBaseResourceTypeIsNotSupported`, but this PR implements it. 

We need need to refactor this to either have a unit test that support QR from population for Tasks or revert the changes specific to this update in order to fix the build.

ellykits: Do we need that check on Line 663?

aurangzaibumer: if subjectType is Patient, we won't be repopulating it twice or if the subjectType is Group, we don't have to populate it for that too. That is why I did add this check.

ellykits: I think we can apply negation instead, then update `loadPopulationResources` to work with other resource types (not just the `Task` resource). My assumption: the function should be used to populate any resource type.

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireViewModel.kt
```diff
@@ -649,12 +650,28 @@ constructor(
...
+    var questionnaireResponse: QuestionnaireResponse? = null
+    // if questionnaireType is Default that means we don't have to load response from DB
+    if (!questionnaireConfig.type.isDefault()) {
+      questionnaireResponse =
+        loadQuestionnaireResponse(subjectId, subjectType, questionnaire.logicalId)
+    }
+    var populationResources = ArrayList<Resource>()
     if (questionnaireResponse == null) {
-      val populationResources = loadPopulationResources(subjectId, subjectType)
+      if (subjectType == ResourceType.Task) {
```
ellykits, line 663:
```suggestion
      if (!subjectType.isIn(ResourceType.Group, ResourceType.Patient) {
```

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireViewModel.kt
ellykits: Suggested refactor for `loadPopulationResources`

```kotlin

 /**
   * Loads resources to be populated into a Questionnaire Response.
   *
   * @param subjectId resource uuid
   * @param subjectType resource type
   */
  private suspend fun loadPopulationResources(
    subjectId: String,
    subjectType: ResourceType
  ): ArrayList<Resource> {
    val populationResources = arrayListOf<Resource>()
    populationResources.add(defaultRepository.loadResource(subjectId, subjectType))
    if (subjectType == ResourceType.Patient) {
        loadRelatedPerson(subjectId)?.run { populationResources.add(this) }
    }
    return populationResources
  }
````


`Patient` resource is a special case here that requires loading of 2 resource types (the `Patient` and `RelatedPerson`) resources.  

Then delete functions `loadTask` and `loadGroup` as they are no longer needed.

pld (approved):

pld (commented):
great work @aurangzaibumer !!

pld (approved):

PR #2342, opened 2023-05-12
Refactor QuestionnaireViewModel

SebaMutuku:
**Description**:
If you look at QuestionnaireViewModel class `saveQuestionnaireResponse`, the removed logic depends on adding params to `LAUNCH_QUESTIONNAIRE` workflow. An example of params is added is     
```
{
   "paramType": "UPDATE_DATE_ON_EDIT",
   "linkId": "b32f67966-d2e4-11ed-afa1-0242ac120002",
   "dataType": "STRING",
  "key": "groupReference",
  "value": "@{groupReference}"
}
```
Remember questionnaire params are meant to be propagated through linkId to `questionnaire` during extraction process.If you look at the logic, it only relies on` "value": "@{groupReference}"`, extracts `resourceType` and `resourceId` from there then uses it to query the resource from db. This happens in series of iterations in the class. This does not use `groupResource` config and  param with type `UPDATE_DATE_ON_EDIT`  is also not propagated to questionnaire-extraction(Checked this on `fhir-resources `)  and hence might be redudant.
Below is a sample `groupResource` config

```
"groupResource": {
              "groupIdentifier": "@{familyLogicalId}",
              "memberResourceType": "Group"
            } 
```

**Intentions of the refactor**
We can as well  extract the `logicalId` and `resourceType` from groupResource and borrow load and update resource logic  initially implemented. Advantage of doing this is it doesn't require loops to update and iterate again through the params loading resources from db. This will do away with `UPDATE_DATE_ON_EDIT` param(s) having to be added on the `questionnaireConfig`
This was discussed with initial implementor @ahsanbhatti49  and will be tested before.


pld:
can you add more details to the description on the goal of this refactor?

ahsanbhatti49:
@SebaMutuku  We need to add 
` "groupResource": {
              "groupIdentifier": "@{familyLogicalId}",
              "memberResourceType": "Group"
            }`

in every menu item from where a group/Resource can be updated.  I have tested Pregnancy outcome and its working find with different scenarios after adding groupResource to default_profile_config menu item.
Also we need to update  UPDATE_DATE_ON_EDIT to PARAMDATA.

SebaMutuku:
Discussed this with @ahsanbhatti49 and @ellykits and the current implementation is okay. Closing this PR

PR #2348, opened 2023-05-12
Clear replay cache once questionnaire submission event is consumed

### LINKED ISSUE #2344: Snackbar message displayed everytime user navigates to household profile after adding a member
**Describe the bug**
After a adding a new family member to a household, the `Household member has been added` snackbar message is displayed everytime user navigates to the household profile.

This snackbar message should only be displayed once after adding the new member

**To Reproduce**
1. Navigate to household register by clicking on one of the families in the household register
2. Click on the fab button on the bottom left of the screen to open the `Add household member` form
3. Fill in the form and submit
4. App navigates back to household profile where a `Household member has been added` snackbar message is displayed
5. Navigate to the household register.
6. Navigate back to the household profile
7. Note that the `Household member has been added` snackbar message is displayed again

**Expected behavior**
The `Household member has been added` snackbar message should only be displayed once after addition of a family member.

 - Device: Google Pixel 2 
 - Android version: Android 11
 - Version 30

Rkareko:

Fixes #2344 

**Engineer Checklist**

**Code Reviewer Checklist**


android/quest/src/main/java/org/smartregister/fhircore/quest/event/EventBus.kt
```diff
@@ -27,4 +27,6 @@ class EventBus @Inject constructor() {
   val events = _events.asSharedFlow()
 
   suspend fun triggerEvent(event: AppEvent) = _events.emit(event)
+
+  fun resetReplayCache() = _events.resetReplayCache()
```
ellykits, line 31: Is it possible to have some sort of a listener to when an event has been consumed? It's easy to forget calling this function.

ellykits (changes requested):

android/quest/src/main/java/org/smartregister/fhircore/quest/event/EventBus.kt
ellykits: This means we have to call it each time an event is consumed which is a bit repetitive

ndegwamartin: Quick one(s), for our use case:
1. Do we expect our events to have multiple subscribers that all need to process it? 
2. Must all the events be processed or it is okay to drop some and pick the latest?
3. Do we intend to support different types of Events from this _sharedFlow_?

ellykits: To answer your questions @ndegwamartin 
1. Yes we may have such use case example listening to sync state events.
2. All event should be processed
3. Yes we already do other than `QuestionnaireSubmission` we will have other event types.

ndegwamartin: Got it, we probably need to factor in ensuring that we initialise all subscribers before any event emissions since the replay cache only stores the last event for late subscribers

ellykits (approved):

ellykits (approved):

PR #2366, opened 2023-05-22
Adding the ability for implementations to define there own regex and separator while joining strings

### LINKED ISSUE #2369: Updating the `RulesEngineService#joinToString()` to allow the regex and separator definition on the configs
**Describe the bug**
- The current `RulesEngineService#joinToString() removes any text that has special characters e.g **CM-NTD**. This happens because the current `RulesEngineService#joinToString()` does not allow the content author to use a custom regex and separator.  
- Overloading the method will allow the configs to use the default functionality that uses the default regex  `"(?<=^|,)[\\s,]*(\\w[\\w\\s]*)(?=[\\s,]*$|,)"` and `", "` separator.  
- The new overloaded method will allow the authors to use define their own regex and separator while authoring the configs. 

**To Reproduce**
1. Assuming you have conditions you show on the profile.
2. Use the `service.joinToString` method on the rules in order to get a comma-separated string.
3. If your initial list contained text with special characters you will notice they won't be returned. 

**Expected behaviour**
- [ ] The `service.joinToString` should return all the required text without removing the text that has special characters.

dubdabasoduba:

Fixes #2369

**Engineer Checklist**

**Code Reviewer Checklist**
- [x] I have verified **Unit tests** have been written for any new feature(s) and edge cases
- [x] I have verified any strings visible on UI components are in the `strings.xml` file
- [x] I have verifed the [CHANGELOG.md](./CHANGELOG.md) file has any notable changes to the codebase
- [x] I have verified the solution has been implemented in a configurable and generic way for reuseable components
- [x] I have built and run the FHIRCore app to verify the change fixes the issue and/or does not break the app


android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
```diff
@@ -387,12 +387,19 @@ constructor(
...
+    }
+
+    /**
+     * This function combines all string indexes to a list separated by the separator defined by the
+     * implementor
+     */
+    fun joinToString(sourceString: MutableList<String?>, regex: String, separator: String): String {
+      sourceString.removeIf { it == null }
+      val inputString = sourceString.joinToString()
+      return regex.toRegex().findAll(inputString).joinToString(separator) { it.groupValues[1] }
```
dubdabasoduba, line 402: @ellykits I decide to overload the method so that the default regex can still be used if a project has no need in providing a new regex

android/quest/src/test/java/org/smartregister/fhircore/quest/StructureMapUtilitiesTest.kt
```diff
@@ -106,7 +106,7 @@ class StructureMapUtilitiesTest : RobolectricTest() {
 
     structureMapUtilities.transform(contextR4, baseElement, structureMap, targetResource)
 
-    Assert.assertEquals(4, targetResource.entry.size)
+    Assert.assertEquals(7, targetResource.entry.size)
```
dubdabasoduba, line 109: The tests were updated to match what we have on FHIR resources

pld (dismissed):

ndegwamartin:
@dubdabasoduba  looks like the **PinViewModelTest.testPinLogin** test is a bit flakey, we can ignore it and track/fix as tech debt

ndegwamartin:
Additionally I think for the implementation to be configurable from the synced configs we still need to extend the implementation to pass and use the those configs

dubdabasoduba:
> @dubdabasoduba looks like the **PinViewModelTest.testPinLogin** test is a bit flakey, we can ignore it and track/fix as tech debt

@ndegwamartin I have added the tech debt issue [here](https://github.com/opensrp/fhircore/issues/2370)

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
```diff
@@ -387,12 +387,19 @@ constructor(
...
-      val regex = "(?<=^|,)[\\s,]*(\\w[\\w\\s]*)(?=[\\s,]*$|,)".toRegex()
-      return regex.findAll(inputString).joinToString(", ") { it.groupValues[1] }
+      return joinToString(source, DEFAULT_REGEX, DEFAULT_STRING_SEPARATOR)
+    }
+
+    /**
+     * This function combines all string indexes to a list separated by the separator and regex
+     * defined by the content author
+     */
+    fun joinToString(sourceString: MutableList<String?>, regex: String, separator: String): String {
```
ellykits, line 399: We we use the default regex and separator constants as default parameters on this function. I think kotlin functions using default params are transformed into overloaded functions in the JVM at runtime and should work well with JEXL, maybe not.

ndegwamartin (approved):

ndegwamartin (approved):

Issue #2393, opened 2023-05-27
[Immunization V2] Only change the dependant Task status if the Pre-requisite has a status `Completed `


dubdabasoduba:
**Describe the bug**
- We currently, have some functionality to update the dependent tasks `Task.executionPeriod.start` and `Task.status` immediately after the Pre-requisite is complete. 
- However, when a workflow is generated the service to update the task statuses seem to run and update the dependent task statuses to DUE and OVERDUE immediately. 
- We need to update this service to make sure it checks for pre-requisites and only updates the dependent tasks `Task.executionPeriod.start` and `Task.status` when the pre-requisite is marked as `Completed`

**To Reproduce**
Steps to reproduce the behaviour:
1. Using [this](https://github.com/opensrp/fhir-resources/pull/1546) PR for configs, 
2. Register a child who is 2 years old. 
3. The Immunization schedule will be generated. 
4. After the `Task.status` update service runs, your will notice all the tasks become DUE or OVERDUE. 

**Expected behaviour**
- [ ] The dependant tasks should not change status from `Requested` unless the Pre-requisites have a status `Completed`

PR #2419, opened 2023-06-06
Fix task scheduling issue

maimoonak:
Fixes https://github.com/opensrp/fhircore/issues/2393 and Task scheduling issue

**Engineer Checklist**

**Code Reviewer Checklist**


android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirTaskExpireWorker.kt
```diff
@@ -40,30 +35,12 @@ constructor(
...
-      while (tasks.isNotEmpty()) {
-        val resultPair = fhirTaskExpireUtil.expireOverdueTasks(lastAuthoredOnDate = maxDate)
-        maxDate = resultPair.first
-        tasks = resultPair.second
-      }
-
-      sharedPreferences.write(
-        SharedPreferenceKey.OVERDUE_TASK_LAST_AUTHORED_ON_DATE.name,
-        maxDate?.formatDate(SDF_YYYY_MM_DD)
-      )
```
ellykits, line 66: Why is this code deleted? This was to act as measure so as not to freeze the application when the background job is running.

android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirTaskPlanWorker.kt
```diff
@@ -58,14 +58,6 @@ constructor(
...
     return withContext(dispatcherProvider.io()) {
-      val appConfig =
-        configurationRegistry.retrieveConfiguration<ApplicationConfiguration>(
-          ConfigType.Application
-        )
-      val batchSize = appConfig.taskBackgroundWorkerBatchSize
-      val lastOffset =
-        sharedPreferencesHelper.read(key = WORK_ID.lastOffset(), defaultValue = "0")!!.toInt()
-
       Timber.i("Running Task status updater worker")
```
ellykits, line 61: Why is this implementation deleted?

android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirTaskExpireUtil.kt
```diff
@@ -58,28 +55,22 @@ constructor(@ApplicationContext val appContext: Context, val fhirEngine: FhirEng
...
-            filter(
-              Task.AUTHORED_ON,
-              {
-                prefix = ParamPrefixEnum.GREATERTHAN_OR_EQUALS
-                value = of(DateTimeType(lastAuthoredOnDate))
-              }
-            )
-          }
-          count = tasksCount
-          sort(Task.AUTHORED_ON, Order.ASCENDING)
```
ellykits, line 71: The authored on was used here to limit the amount of data to process at a time, otherwise a pre-configured batch size would be used. This background process is expensive and blocking especially during the database reads.

ellykits (changes requested):

android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirTaskExpireWorker.kt
ellykits: How is this related to the tagged issue?

android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirTaskPlanWorker.kt
```diff
@@ -58,14 +58,6 @@ constructor(
...
   override suspend fun doWork(): Result {
     return withContext(dispatcherProvider.io()) {
-      val appConfig =
-        configurationRegistry.retrieveConfiguration<ApplicationConfiguration>(
-          ConfigType.Application
-        )
-      val batchSize = appConfig.taskBackgroundWorkerBatchSize
-      val lastOffset =
-        sharedPreferencesHelper.read(key = WORK_ID.lastOffset(), defaultValue = "0")!!.toInt()
-
```
Rkareko, line 68: The logic for batching was introduced in this [PR](https://github.com/opensrp/fhircore/pull/2351) as part of the performance enhancements for background workers.
Do you mind explaining why it has been removed?

maimoonak:
@ellykits The storage of last process task (authoredDate/any other id) would lead to failed updates for all tasks in future once they are processed. A task created today would need to be checked again and again in future if it should be expired or made ready until it has been cancelled/completed. 

I have changed the implementation to use task.start/task.end (Task.period) to pick tasks which should be expired OR made ready. This already is now leading a very small subset of tasks (Unless app was never used and now leading to huge number for first time processing). If we still need to make it more performant we would want to add a service on server that ends replicates this functionality. In an ideal case by new logic there should not be a large number tasks that are to be expired today OR made ready for processing.

android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirTaskPlanWorker.kt
ellykits: Can you respond to @Rkareko question above? Will the refactored filter improve the query performance? I think regardless of the refactor done on the filters, the query will still run against all the available Tasks in the app.

maimoonak: The result of the query for Tasks is not predictable and not always same amount of tasks and hence we can not use the offset/batch size broken down across different executions. There are other Jobs which mark tasks expired as well, or some times tasks can also be completed by form submissions, and we can in future have a functionality that moves task from one state to another. A task itself can have or be eligible for a different state at different point of time. Lets say today we have 1000 tasks which resulted with given params in query with batch size of 50. We would update 50 tasks and next might be updated after 1 hour, or maybe 2 hours when this is executed. But in the mean time 100 Tasks were expired by other service, 100 were completed by form submission or whatso ever. Now when after an hour query runs the tasks would be 800 and offset starts from 50. We have no way to tell what Tasks we have missed due to sorting order, and status updates. Also as opposed to getting ALL tasks of given statuses, we are now filter tasks with status and due date which would be lesser in  number, hence making lesser impact on performance. However, if still we want to improve performance the Task query count and objects should be predictable, and with same entities and in same order

dubdabasoduba:
@ellykits @Rkareko @maimoonak here are the Task statuses as defined. 

**Tasks status**
- **Requested** -- This status is given to a Task when the `executionPeriod.start` date is in the future. 
- **Ready** -- This status is given to a Task when the `executionPeriod.start` date is <= to today and the `executionPeriod.end` date >= today
    - The background task should change the Task status to `ready` if the `executionPeriod.start` date is <= to today and the `executionPeriod.end` date >= today
- **In Progress** -- This status is given to a Task when the QuestionnaireResponse it acts on is marked as `draft`
- **Completed** -- This status is given to a Task when the QuestionnaireResponse it acts on is marked as `completed`
- **Overdue** -- Is status is not assigned to the Task resource. This status is inferred from the following logic. A **Ready** / **In Progress** task that the `executionPeriod.end` date < today
   - Since this is not an actual status we might want to write a function on the `RuleFactory` to get the status to OVERDUE in order to change the Task Background color. 
- **Expired** -- This status will be represented by the `cancelled` status on the Task resource. This is a Task that has been Overdue for a while and therefore it can't be performed anymore. 
   - We have 2 ways to achieve this. 
   - It can be achieved through the Workflow closure from the Event Management functionality.  
   - It can also be closed when the date today > `Restriction.period.end`

ellykits:
So that you know – the discussion already happened. I have removed the unnecessary labels.

ellykits:
@maimoonak To share a few stats to demonstrate any performance gains from the new queries (NOT required to merge this PR though)

ellykits (approved):

android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirTaskPlanWorker.kt
ellykits: Resolved this conversation in a separate meeting. We confirmed @maimoonak's observation above.

PR #2515, opened 2023-07-03
add configuration to generate MeasureReports in the background

roywanyaga:
**IMPORTANT: Where possible all PRs must be linked to a Github issue**

Closes #[2307]

**Engineer Checklist**
- [ ] I have written **Unit tests** for any new feature(s) and edge cases for bug fixes
- [ ] I have added any strings visible on UI components to the `strings.xml` file
- [ ] I have updated the  [CHANGELOG.md](./CHANGELOG.md) file for any notable changes to the codebase
- [ ] I have run `./gradlew spotlessApply` and `./gradlew spotlessCheck` to check my code follows the project's style guide
- [ ] I have built and run the FHIRCore app to verify my change fixes the issue and/or does not break the app 
- [ ] I have checked that this PR does NOT introduce **breaking changes** that require an update to **_Content_** and/or **_Configs_**? _If it does add a sample here or a link to exactly what changes need to be made to the content._

**Code Reviewer Checklist**
- [ ] I have verified **Unit tests** have been written for any new feature(s) and edge cases
- [ ] I have verified any strings visible on UI components are in the `strings.xml` file
- [ ] I have verifed the [CHANGELOG.md](./CHANGELOG.md) file has any notable changes to the codebase
- [ ] I have verified the solution has been implemented in a configurable and generic way for reuseable components
- [ ] I have built and run the FHIRCore app to verify the change fixes the issue and/or does not break the app

PR #2520, opened 2023-07-04
Add configuration to generate MeasureReports in the background (#2515)

### LINKED ISSUE #207: Notice D - mADX: Activity 1.2.1.1: Develop a sync mechanism of FHIR MeasureReport from Android to the HMIS.

### LINKED ISSUE #2307: [In-App Reporting] Add configuration to enable loading of specific reports in the background
Update measure report configuration to allow loading of specific reports in the background. 
The configuration will include:

- [ ] An indication of whether the report is to be loaded in the background
- [ ] ISO Period indicating the frequency of the background job

Research to ensure that the Measure report is generated in the background without affecting users

**Additional context**
An implementation to load all in-app reports in the background had been done but disabled because it was affecting performance and application use;
As CQL evaluation is being reviewed to enhance performance this will be an initial solution to enable loading of specific reports in the background

**Acceptance criteria**

- [ ] User update a specific measure report config to enable loading in the background
- [ ] Measure report is generated in the background without affecting users
- [ ] ~Whenever a background task is running show a message in the UI so the user knows the reason if the UX lags~
    - should this be a non-dismissable toast?
    - should this show for all background tasks regardless, or only ones that are configured to run on a schedule?

ndegwamartin:

Fixes #2307

**Engineer Checklist**

**Code Reviewer Checklist**


pld:
@ndegwamartin this doesn't seem like the correct fixes ref

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/MeasureReportViewModel.kt
pld, line 1: nice! changes to this class show why this is such a useful refactor!

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/worker/MeasureReportConfigWorker.kt
```diff
@@ -0,0 +1,109 @@
...
+            measureReportRepository.evaluatePopulationMeasure(
+              config.url,
+              startDateFormatted,
+              endDateFormatted,
+              subjects,
+              existing
+            )
+          }
+        }
+      Timber.i("Result.success  / . . . MeasureReportWorker . . ./")
```
pld, line 96: same thing
```suggestion
      Timber.i("successfully completed MeasureReportWorker")
```

pld (changes requested):

android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/report/measure/MeasureReportConfiguration.kt
```diff
@@ -26,8 +26,9 @@ data class MeasureReportConfiguration(
   override var configType: String = ConfigType.MeasureReport.name,
   val id: String,
   val registerId: String,
-  val registerDate: String? = null,
+  val startPeriod: String? = "2023-01-01",
```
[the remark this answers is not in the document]
roywanyaga: This particular field represents the start date for the in-app reports. The list of months will show as back as this date. We thought renaming it to `startPeriod `would be ideal instead of `registerDate`. If left to null and if it's missed in the config the application will crash.

pld:
Ok then we should use a constant for now and add a code comment explaining it in case we need to move it to load from a config later

android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/report/measure/MeasureReportConfiguration.kt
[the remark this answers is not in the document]
pld: Ok then we should use a constant for now and add a code comment explaining it in case we need to move it to load from a config later

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/worker/MeasureReportConfigWorker.kt
```diff
@@ -0,0 +1,109 @@
...
+
+      inputData
+        .getString(MEASURE_REPORT_CONFIG_ID)
+        ?.let<String, MeasureReportConfiguration> {
+          configurationRegistry.retrieveConfiguration(ConfigType.MeasureReport, it)
+        }
+        ?.reports
+        ?.forEach { config ->
+          val startDateFormatted = today().firstDayOfMonth().formatDate(SDF_YYYY_MM_DD)
+          val endDateFormatted = today().lastDayOfMonth().formatDate(SDF_YYYY_MM_DD)
```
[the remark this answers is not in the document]
pld: A docstring note that descriptions the assumptions about monthly periods and needs to run once a day should be sufficient

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/report/measure/worker/MeasureReportMonthPeriodWorker.kt
```diff
@@ -0,0 +1,113 @@
...
+  val defaultRepository: DefaultRepository,
+  val dispatcherProvider: DispatcherProvider,
+  private val measureReportRepository: MeasureReportRepository,
+  val configurationRegistry: ConfigurationRegistry,
+) : CoroutineWorker(context, workerParams) {
+
+  /**
+   * This method always sets the start and end date of the month to be the first and last day of the
+   * current month. If you are using this worker to generate monthly reports it enqueue a new worker
+   * at least every 24 hours.
```
pld, line 60: I'm still not sure this is correct, need to think about this a bit more to see if there's an edge case that misses data from the previous month when the date changes.

pld: I think there is actually still a problem, since it’s using “today” to figure out the current month instead of the date it was enqueued. This is what I’m thinking, if it’s the last day of the month, you run the worker, and enqueue another worker for tomorrow. Then you collect more data. That new data you collected will never appear in the previous months measure report, because by the time the enqueued measure report runs it will tomorrow, in the next month, and that run will be for the next month.

pld: One (simple?) solution is to have `today` be set when the worker is enqueued, instead of when the worker is run. then when you enqueue a worker on the last day of the month, and proceed to collect some data, when that worker runs it will cover all the data for the previous month, during which it was enqueued.

This also seems to match expected behavior... better?

pld: created this issue to track this, https://github.com/opensrp/fhircore/issues/2567

pld (approved):
follow-up tasks in https://github.com/opensrp/fhircore/issues/2567

PR #2523, opened 2023-07-04
Fix refreshing profile data

ellykits:

Fixes https://github.com/opensrp/fhir-resources/issues/1940

**Engineer Checklist**

**Code Reviewer Checklist**

### DIFF
--- a/android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirTaskUtil.kt
+++ b/android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirTaskUtil.kt
@@ -30,22 +30,24 @@ import org.hl7.fhir.r4.model.DateTimeType
 import org.hl7.fhir.r4.model.ResourceType
 import org.hl7.fhir.r4.model.Task
 import org.smartregister.fhircore.engine.data.local.DefaultRepository
+import org.smartregister.fhircore.engine.util.extension.executionStartIsBeforeOrToday
 import org.smartregister.fhircore.engine.util.extension.expiredConcept
 import org.smartregister.fhircore.engine.util.extension.extractId
+import org.smartregister.fhircore.engine.util.extension.isIn
 import org.smartregister.fhircore.engine.util.extension.isPastExpiry
+import org.smartregister.fhircore.engine.util.extension.plusDays
 import org.smartregister.fhircore.engine.util.extension.toCoding
 import timber.log.Timber
 
 @Singleton
-class FhirTaskExpireUtil
+class FhirTaskUtil
 @Inject
 constructor(@ApplicationContext val appContext: Context, val defaultRepository: DefaultRepository) {
 
   /**
    * Fetches and returns tasks whose Task.status is either "requested", "ready", "accepted",
    * "in-progress" and "received" and Task.restriction.period.end is <= today. It uses the maximum .
-   * The size of the tasks is between 0 to (tasksCount * 2). It is not guaranteed that the list of
-   * tasks returned will be of size [tasksCount].
+   * The size of the tasks is between 0 to (tasksCount * 2).
    */
   suspend fun expireOverdueTasks(): List<Task> {
     Timber.i("Fetch and expire overdue tasks")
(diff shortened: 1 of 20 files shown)


ellykits:
### Breaking config changes 

#### Removed
property: "`refreshContent`"
Affected config: `QuestionnaireConfig`

#### Added
Affected config: `QuestionnaireConfig`
New property: `onSubmitActions` 
New property type: `List<ActionConfig>`

Sample
```json
             "onSubmitActions": [
                  {
                    "trigger": "ON_QUESTIONNAIRE_SUBMISSION",
                    "workflow": "LAUNCH_REGISTER",
                    "id": "householdRegister",
                    "display": "All Households",
                    "popNavigationBackStack": true
                  }
            ]
```

android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirTaskUtil.kt
```diff
@@ -108,4 +110,58 @@ constructor(@ApplicationContext val appContext: Context, val defaultRepository:
...
+      ) {
+        Timber.i("Task ${task.id} marked ready")
+        task.status = Task.TaskStatus.READY
+        defaultRepository.update(task)
+      }
+    }
+  }
+
+  /**
+   * Check in the task is part of another task and if so check if the parent task is
```
pld, line 150:
```suggestion
   * Check if the task is part of another task and if so check if the parent task is
```

android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirTaskUtil.kt
```diff
@@ -108,4 +110,58 @@ constructor(@ApplicationContext val appContext: Context, val defaultRepository:
...
+   * completed,cancelled,failed or entered in error.
+   */
+  private suspend fun Task.preReqConditionSatisfied() =
```
pld, line 153: I think we can rename this fun and consolidate the precondition

```suggestion
  private suspend fun Task.preconditionSatisfied() =
      this.executionStartIsBeforeOrToday() &&
      this.status == Task.TaskStatus.REQUESTED &&
```

android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirTaskUtil.kt
```diff
@@ -108,4 +110,58 @@ constructor(@ApplicationContext val appContext: Context, val defaultRepository:
...
+      }
+
+    Timber.i("Found ${tasks.size} tasks to be updated")
+
+    tasks.forEach { task ->
+      // expired tasks are handled by other service i.e. FhirTaskExpireWorker
+      if (task.executionStartIsBeforeOrToday() &&
+          task.status == Task.TaskStatus.REQUESTED &&
+          task.preReqConditionSatisfied()
+      ) {
```
pld, line 141: why not also consider the above 2 conditions part of the precondition, and add them to that function?

```suggestion
      if task.preReqConditionSatisfied() {
```

android/engine/src/main/java/org/smartregister/fhircore/engine/util/extension/ResourceExtension.kt
```diff
@@ -323,17 +323,13 @@ fun String.resourceClassType(): Class<out Resource> =
 fun String.extractLogicalIdUuid() = this.substringAfter("/").substringBefore("/")
 
 /**
- * You provide a suspended function in Kotlin, which updates the due date of a task's dependent
- * tasks based on the date of a related immunization. The function takes a [defaultRepository]
- * parameter that is an instance of [DefaultRepository]. It then loops through all the tasks that
- * this task is a part of, loads the dependent tasks and their related immunization resources from
- * the repository, and updates the start date of the dependent task if it's scheduled to start
+ * This suspend function in updates the due date of the dependents of the current [Task], based on
```
pld, line 326:
```suggestion
 * This suspend function updates the due date of the dependents of the current [Task], based on
```

android/engine/src/main/java/org/smartregister/fhircore/engine/util/extension/ResourceExtension.kt
```diff
@@ -323,17 +323,13 @@ fun String.resourceClassType(): Class<out Resource> =
...
+ * the date of a related [Immunization] [Task]. The function loops through all the tasks that are
+ * part-of the current, loads the dependent tasks and their related immunization resources from the
```
pld, line 328:
```suggestion
 * part-of the current task, loads the dependent tasks and their related immunization resources from the
```

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/profile/ProfileFragment.kt
```diff
@@ -97,34 +98,44 @@ class ProfileFragment : Fragment() {
   override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
     viewLifecycleOwner.lifecycleScope.launch {
       viewLifecycleOwner.repeatOnLifecycle(Lifecycle.State.CREATED) {
+        // Each profile should have unique eventId
```
pld, line 101:
```suggestion
        // Each profile should have a unique eventId
```

pld (commented):
lgtm! test failure looks related to missing fun call mock on eventBus

dubdabasoduba:
> @dubdabasoduba before we split the PR for the 0.2.4 branch could you please confirm that the logic for updating task status is correct?
> 
> From the logic seems like we:
> 
> 1. Look for any tasks that are due with a task status of `REQUESTED`
> 2. Look for the parent task they are a `partOf`
> 3. If the parent task is found and it as the status _CANCELLED, COMPLETE, FAILED_ or _ENTEREDINERROR_ then we mark the (child)task status as `READY`
> 
> **Note**: The above logic was not added by this PR just refactored into it.
> 
> Also please review the code to confirm my assertions above.

@ndegwamartin this is correct. We might want to check in with the PMs on enabling dependent tasks when the pre-requisite was CANCELLED, FAILED or ENTEREDINERROR

dubdabasoduba:
@ndegwamartin this works well. 
- The above config allows redirecting to a register. 
    - It however breaks with redirecting to a profile. 
- The Profile refreshes correctly.   

This [PR](https://github.com/opensrp/fhircore/pull/2532) is also needed to avoid a crash on remove family functionality

ellykits:
> @ndegwamartin this works well.
> 
> * The above config allows redirecting to a register.
>   
>   * It however breaks with redirecting to a profile.
> * The Profile refreshes correctly.
> 
> This [PR](https://github.com/opensrp/fhircore/pull/2532) is also needed to avoid a crash on remove family functionality

Could you let me know which profile you tried opening here? if the profile is related to a household it would work with the `Group` resource being the base. Otherwise, the `ResourceData` would be null (`ResourceData` is required to open the profile).

android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirTaskUtil.kt
ellykits: I see. I think the conditions here are breaking the functionality for marking Tasks as ready. The conditions for updating the dependent Tasks should be separate from the parent Task.

ndegwamartin (approved):

ellykits:
@dubdabasoduba I deleted the conditional checks that relied on the dependent Tasks, for which I needed confirmation from you.

pld (approved):

PR #2584, opened 2023-07-13
Pld doc updates

rowo:
@pld did I do this correctly?


pld:
noop, but I'll handle it, please don't make any changes on this branch until I let you know...

PR #2599, opened 2023-07-18
Fixes the app showing a "Sync Complete" message even on sync failure

### LINKED ISSUE #2575: The app shows a "Sync Complete" toaster even if sync failed
**Describe the bug**
- Currently, the application would emit a _Sync Complete_ toaster even if the sync didn't complete successfully. 
- This is confusing to the users and might lead to data loss in case users deleted the app based on the toaster. 

**To Reproduce**
1. Go to the Application
2. Generate some invalid data that can't sync to the server. 
3. Click on the sync icon 
4. Wait and see the response after it fails

**Expected behaviour**
- [ ] _Sync Complete_ Should only be emitted if all the required data has been synced to and from the sync.

ndegwamartin:

Closes #2575 

**Additionally Includes:**
- Fix Missing Manifest Resources on Setup
- Enhance Sync Progress View logic
- Fix NON PROXY flag incorrect state in dev mode
- Improve UX when launching Insights

**Engineer Checklist**

**Code Reviewer Checklist**


ellykits (approved):

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/appsetting/AppSettingViewModel.kt
```diff
@@ -129,15 +132,20 @@ constructor(
...
+                        profileConfig
+                          ?.fhirResource
+                          ?.dependentResourceTypes(
+                            patientRelatedResourceTypes,
+                          )
+                      } else {
+                        registerConfig.fhirResource.dependentResourceTypes(
+                          patientRelatedResourceTypes,
+                        )
+                      }
```
pld, line 148: because we are ignoring unknown keys, as it was written it always decodes as RegisterConfiguration

ndegwamartin: with the new implementation looks like we might experience a crash if registerConfig was actually ever null

pld: maybe, it would be a pre-existing issue since that's the same as config getting set to null in the previous version, I think we have a test for that, but it might be passing b/c of mocks

PR #2699, opened 2023-08-19
OpenSRP 0.2.5 v1

ndegwamartin:

Fixes https://github.com/onaio/fhir-resources/issues/2531

### DIFF
--- a/android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirCarePlanGenerator.kt
+++ b/android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirCarePlanGenerator.kt
@@ -170,7 +170,7 @@ constructor(
 
     if (carePlanModified) saveCarePlan(output)
 
-    fhirTaskUtil.updateTaskStatuses()
+    fhirTaskUtil.updateTaskStatuses(subject)
 
     return if (output.hasActivity()) output else null
   }
--- a/android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirCarePlanGenerator.kt
+++ b/android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirCarePlanGenerator.kt
@@ -168,9 +168,12 @@ constructor(
       }
     }
 
+    val careplanTasks =
+      output.contained.filter { it.resourceType == ResourceType.Task }.map { it as Task }
+
     if (carePlanModified) saveCarePlan(output)
 
-    fhirTaskUtil.updateTaskStatuses(subject)
+    fhirTaskUtil.updateTaskStatuses(careplanTasks)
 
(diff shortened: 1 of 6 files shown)


dubdabasoduba (approved):
@ndegwamartin This works to reduce the number of resources to update. We can maybe update it further to target the specific Tasks created from the extraction.

ndegwamartin:
> @ndegwamartin This works to reduce the number of resources to update. We can maybe update it further to target the specific Tasks created from the extraction.

@dubdabasoduba Yeah we can do that. Is the performance acceptable for the release so that we can track it on a separate ticket or would the update be a significant performance upgrade? Also wouldn't we want to have all the users tasks updated in real time for the current patient of interest?

dubdabasoduba:
> > @ndegwamartin This works to reduce the number of resources to update. We can maybe update it further to target the specific Tasks created from the extraction.
> 
> @dubdabasoduba Yeah we can do that. Is the performance acceptable for the release so that we can track it on a separate ticket or would the update be a significant performance upgrade? Also wouldn't we want to have all the users tasks updated in real time for the current patient of interest?

It's definitely going to reduce the number of tasks evaluated further. The workflow with the highest number of Tasks has 60. The one that follows has 10 tasks. Which means if you are performing any other workflow you are going to have less than 10 tasks to evaluate

dubdabasoduba:
> > > @ndegwamartin This works to reduce the number of resources to update. We can maybe update it further to target the specific Tasks created from the extraction.
> > 
> > 
> > @dubdabasoduba Yeah we can do that. Is the performance acceptable for the release so that we can track it on a separate ticket or would the update be a significant performance upgrade? Also wouldn't we want to have all the users tasks updated in real time for the current patient of interest?
> 
> It's definitely going to reduce the number of tasks evaluated further. The workflow with the highest number of Tasks has 60+. The one that follows has 10 tasks. This means if you are performing any other workflow you are going to have less than 10 tasks to evaluate

dubdabasoduba (approved):

ndegwamartin:
> > @dubdabasoduba Yeah we can do that. Is the performance acceptable for the release so that we can track it on a separate ticket or would the update be a significant performance upgrade? Also wouldn't we want to have all the users tasks updated in real time for the current patient of interest?
> 
> It's definitely going to reduce the number of tasks evaluated further. The workflow with the highest number of Tasks has 60+. The one that follows has 10 tasks. This means if you are performing any other workflow you are going to have less than 10 tasks to evaluate
> 
> As agreed let's make the update on the PR to `main`

Actually the 70+ tasks per submission feels like a significant number. My understanding was that they'd always be less than 10 during the evaluation. The LOE might be small, let me see

dubdabasoduba (approved):

PR #2740, opened 2023-09-08
Fix the SQL error throw when filtering

dubdabasoduba:

Fixes https://github.com/onaio/fhir-resources/issues/2635

**Engineer Checklist**

**Code Reviewer Checklist**

### DIFF
--- a/android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirCarePlanGenerator.kt
+++ b/android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirCarePlanGenerator.kt
@@ -171,10 +171,12 @@ constructor(
 
     if (carePlanModified) saveCarePlan(output)
 
-    fhirResourceUtil.updateUpcomingTasksToDue(
-      subject = subject.asReference(),
-      taskResourcesToFilterBy = carePlanTasks,
-    )
+    if (carePlanTasks.isNotEmpty()) {
+      fhirResourceUtil.updateUpcomingTasksToDue(
+        subject = subject.asReference(),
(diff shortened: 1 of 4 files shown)


android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirCarePlanGenerator.kt
```diff
@@ -171,10 +171,12 @@ constructor(
...
-    fhirResourceUtil.updateUpcomingTasksToDue(
-      subject = subject.asReference(),
-      taskResourcesToFilterBy = carePlanTasks,
-    )
+    if (carePlanTasks.isNotEmpty()) {
+      fhirResourceUtil.updateUpcomingTasksToDue(
+        subject = subject.asReference(),
+        taskResourcesToFilterBy = carePlanTasks,
+      )
+    }
```
pld, line 179: is this for efficiency reasons? If we did call it, would it not be a problem, but it would always be a no-op?

Wondering if we should push this check into the function

dubdabasoduba: Yah, I think we have no reason to call the method if the task list is empty.

pld: OK after reading `updateUpcomingTasksToDue` it does not make sense to push this check in and I get why this is a special case since it's being called only after tasks just generated. I also can't see an obvious way to test this now

pld (approved):

PR #2746, opened 2023-09-13
Implement an optional and configurable code path that uses $apply for PlanDefenition execution

### LINKED ISSUE #2733: Implement an optional and configurable code path that uses $apply for PlanDefenition execution
As a content creator and project implementer I want to be able to run $apply on a PlanDefinition and any referenced CQL to create output resources, e.g. CarePlans.

I want to be able to optionally enable this instead of, or in addition to, the current method of executing PlanDefinitions.

**Additional context**
Our current approach to executing PlanDefinitions is to run a StructureMap on the PlanDefinition that uses FHIRPath within it to create additional resources.

The android-fhir repo has an [example app](https://github.com/google/fhir-app-examples/tree/main/configurablecare) that executes a [PlanDefinition](https://github.com/google/fhir-app-examples/blob/main/configurablecare/app/src/main/assets/fhir-resources-example.json) using $apply through their [Workflow API](https://github.com/google/fhir-app-examples/blob/main/configurablecare/app/src/main/java/com/google/fhir/examples/configurablecare/care/CarePlanManager.kt#L138).

**Acceptance criteria**

- [ ] Google's example PlanDefinition is loaded into an app config that also includes a config to execute it through the app UI
- [ ] Executing the PlanDefinition through the app UI produces the expected output

**Area path**

1. Login to the WHO EIR
2. Open _Navigation bar_
3. Click on _Patient register_
4. Click on _Add a patient_
5. Click on _Run plan def_ configured in the add patient questionnaire

- [ ] Introduce a new boolean config property called `QuestionnaireConfig.generateCarePlanWithWorkflowApi`. When set to `true`, use $apply from the Workflow library to generate CarePlan otherwise default to using the current approach in `FhirCarePlanGenerator#generateOrUpdateCarePlan()`. 

- [ ] Update the signature for the function `FhirCarePlanGenerator#generateOrUpdateCarePlan()` to include a new parameter `generateCarePlanWithWorkflowApi`. The value for this param should be obtained from the `QuestionnaireConfig`, which is available in the `QuestionnaireViewModel` (where the generate CarePlan function is called).

- [ ] NOTE: The `FhirCarePlanGenerator#generateOrUpdateCarePlan()` is an overloaded function. The one to be updated has `planDefinitionId: String` in its parameters and is invoked directly in the `QuestionnaireViewModel`.

- [ ] Implement functionality to execute $apply to generate CarePlan. Ideally, this should be done in a separate class (`WorkflowCarePlanGenerator`). The class will be injected into the `FhirCarePlanGenerator` constructor. 

- [ ] Update the implementation inside `FhirCarePlanGenerator#generateOrUpdateCarePlan()` to either use the new `WorkflowCarePlanGenerator.generateCarePlan()` or the current depending on the value passed through the `QuestionnaireConfig.generateCarePlanWithWorkflowApi` property.

- [ ] The implementation for `WorkflowCarePlanGenerator.generateCarePlan()` can be adapted from this function [applyPlanDefinitionOnPatient](https://github.com/google/fhir-app-examples/blob/e6db1370e9734872472153292cdc1ee2cbed06a2/configurablecare/app/src/main/java/com/google/fhir/examples/configurablecare/care/CarePlanManager.kt#L126C44-L126C44) in the sample application. Copy and modify all the necessary functions from the sample code to `WorkflowCarePlanGenerator` as required. 

- [ ] Remember to document the functions and use the HILT injected instances of the [KnowledgeManager](https://github.com/opensrp/fhircore/blob/c6e17d3dcc2520beeaa930f48bb23c7134b790c3/android/engine/src/main/java/org/smartregister/fhircore/engine/di/CoreModule.kt#L63) and [FhirOperator](https://github.com/opensrp/fhircore/blob/c6e17d3dcc2520beeaa930f48bb23c7134b790c3/android/engine/src/main/java/org/smartregister/fhircore/engine/di/CoreModule.kt#L70). The two can be injected as constructor properties of the `WorkflowCarePlanGenerator` class.

- [ ] Update the tests for `FhirCarePlanGenerator#generateOrUpdateCarePlan()` and write new tests for `WorkflowCarePlanGenerator.generateCarePlan()`. Avoid using mocks unless really necessary, test the actual implementation instead, using injected dependencies.

owais-vd:

Fixes #2733 

**Engineer Checklist**

**Code Reviewer Checklist**


android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirCarePlanGenerator.kt
```diff
@@ -111,71 +117,80 @@ constructor(
...
+              source.setParameter(ActivityDefinition.SP_VERSION, IntegerType(index))
+
+              val structureMap = fhirEngine.get<StructureMap>(IdType(action.transform).idPart)
+              structureMapUtilities.transform(
+                transformSupportServices.simpleWorkerContext,
+                source,
+                structureMap,
+                output,
+              )
+            }
```
pld, line 154: can we put this entire branch path in a `generateOrUpdateCarePlanFromStructureMap` function?

whoops I guess not just these lines but all the way down, this entire `else` branch

android/engine/src/main/java/org/smartregister/fhircore/engine/task/WorkflowCarePlanGenerator.kt
```diff
@@ -0,0 +1,357 @@
...
+      }
+    }
+    return bundleCollection
+  }
+
+  /**
+   * Knowledge resources are loaded from [FhirEngine] and installed so that they may be used when
+   * running $apply on a [PlanDefinition]
+   */
+  private suspend fun loadCarePlanResourcesFromDb() {
```
pld, line 104:
```suggestion
  private suspend fun loadPlanDefinitionResourcesFromDb() {
```

android/engine/src/main/java/org/smartregister/fhircore/engine/task/WorkflowCarePlanGenerator.kt
```diff
@@ -0,0 +1,357 @@
...
+    if (cqlLibraryIdList.isEmpty()) {
+      loadCarePlanResourcesFromDb()
+    }
+
+    val carePlanProposal =
+      fhirOperator.generateCarePlan(planDefinitionId = planDefinitionId, patientId = patientId)
+        as CarePlan
+
+    // Fetch existing CarePlan of record for the Patient or create a new one if it does not exist
+    // val carePlanOfRecord = getCarePlanOfRecordForPatient(patient)
```
pld, line 143: do we need to call this?

pld (commented):
this is great!

android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirCarePlanGenerator.kt
owais-vd: done

android/engine/src/main/java/org/smartregister/fhircore/engine/task/WorkflowCarePlanGenerator.kt
owais-vd: done

owais-vd:
@maimoonak I need your help to understand more about that pls review a PR once you have time. Thanks

android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirCarePlanGenerator.kt
```diff
@@ -148,7 +178,8 @@ constructor(
                 .let { fhirPathEngine.evaluate(null, input, planDefinition, subject, it) }
                 ?.takeIf { it.isNotEmpty() }
                 ?.let { evaluatedValue ->
-                  // TODO handle cases where we explicitly need to set previous value as null, when
+                  // TODO handle cases where we explicitly need to set previous value as null,
+                  // when
                   // passing null to Terser, it gives error NPE
```
pld, line 183:
```suggestion
                  // when passing null to Terser, it gives error NPE
```

android/engine/src/main/java/org/smartregister/fhircore/engine/task/WorkflowCarePlanGenerator.kt
```diff
@@ -0,0 +1,329 @@
...
+   * resources required to $apply a [PlanDefinition] are present within `PlanDefinition.contained`
+   *
+   * @param planDefinition PlanDefinition resource for which dependent resources are extracted
+   */
+  suspend fun getPlanDefinitionDependentResources(
+    planDefinition: PlanDefinition,
+  ): Collection<Resource> {
+    var bundleCollection: Collection<Resource> = mutableListOf()
+
+    for (resource in planDefinition.contained) {
```
maimoonak, line 83: the contained resources might be handled by PlanProcessor within code, however, the resources which are not in contained like ActivityDefintion in WHO plans, Library referenced by PlanDef might need to be part of KnowledgeManager. Kindly double check.

android/engine/src/main/java/org/smartregister/fhircore/engine/task/WorkflowCarePlanGenerator.kt
```diff
@@ -0,0 +1,329 @@
...
+      "CommunicationRequest" -> TODO("Not supported yet")
+      "RequestGroup" -> {}
+      else -> TODO("Not a valid request resource")
+    }
+  }
+
+  /** Update [Task] status */
+  suspend fun updateRequestResourceStatus(resource: Task, status: String) {
+    resource.status = Task.TaskStatus.valueOf(status)
+    resource.executionPeriod.end = Date.from(Instant.now())
```
maimoonak, line 302: whether Task ends or not depends on task status. For our current flow it might be fine but task statuses could intermediary or on-hold as well which means that pause it for now but it is not ended can be resumed later on

maimoonak (commented):
Thanks @owais-vd this looks good. Just double check two comments.

android/engine/src/main/java/org/smartregister/fhircore/engine/task/WorkflowCarePlanGenerator.kt
dubdabasoduba: @owais-vd could we reuse the already existing Task status update functionality?  That way this is handled in the same place throughout the app

android/engine/src/main/java/org/smartregister/fhircore/engine/task/WorkflowCarePlanGenerator.kt
```diff
@@ -0,0 +1,329 @@
...
+      Task.TaskStatus.CANCELLED -> CarePlan.CarePlanActivityStatus.CANCELLED
+      Task.TaskStatus.INPROGRESS -> CarePlan.CarePlanActivityStatus.INPROGRESS
+      Task.TaskStatus.ONHOLD -> CarePlan.CarePlanActivityStatus.ONHOLD
+      Task.TaskStatus.FAILED -> CarePlan.CarePlanActivityStatus.STOPPED
+      Task.TaskStatus.COMPLETED -> CarePlan.CarePlanActivityStatus.COMPLETED
+      Task.TaskStatus.ENTEREDINERROR -> CarePlan.CarePlanActivityStatus.ENTEREDINERROR
+      Task.TaskStatus.NULL -> CarePlan.CarePlanActivityStatus.NULL
+      else -> CarePlan.CarePlanActivityStatus.NULL
+    }
+  }
```
dubdabasoduba, line 328: I might be missing some but why are we doing this?  A CarePlan can have multiple tasks so it's status is not mapped 1 to 1 to a task.

jingtang10:
Very interesting work @owais-vd! We're working on finalising some of the request API design in the workflow library and in my view they can really simplify a lot of the `switch`/`when` statements in this PR. I think it would make a lot of sense for this type of logic to be maintained in a separate library and we can maintain that logic against the standard.

@pld @dubdabasoduba fyi

android/engine/src/main/java/org/smartregister/fhircore/engine/task/WorkflowCarePlanGenerator.kt
owais-vd: https://github.com/opensrp/fhircore/blob/e7860d69e8766b4caa33c0e8b82b0bcb03001d51/android/engine/src/main/java/org/smartregister/fhircore/engine/task/WorkflowCarePlanGenerator.kt#L151-L165

creating new activity on each task

android/engine/src/main/java/org/smartregister/fhircore/engine/task/WorkflowCarePlanGenerator.kt
```diff
@@ -0,0 +1,322 @@
...
+   *
+   * @param planDefinitionId PlanDefinition resource ID for which $apply is run
+   * @param patient Patient resource for which the [PlanDefinition] $apply is run
+   * @param requestResourceConfigs List of configurations that need to be applied to the request
+   *   resources as a result of the proposed [CarePlan]
+   */
+  suspend fun applyPlanDefinitionOnPatient(
+    planDefinition: PlanDefinition,
+    patient: Patient,
+    data: Bundle = Bundle(),
```
pld, line 123: this doesn't appear to be used

android/engine/src/main/java/org/smartregister/fhircore/engine/task/WorkflowCarePlanGenerator.kt
```diff
@@ -0,0 +1,322 @@
...
+    val carePlanProposal =
+      fhirOperator.generateCarePlan(planDefinitionId = planDefinitionId, patientId = patientId)
+        as CarePlan
+
+    // Accept the proposed (transient) CarePlan by default and add tasks to the CarePlan of record
+    acceptCarePlan(carePlanProposal, output)
+  }
+
+  /** Update the [CarePlan] to include a reference to the FHIR-define protocol or guideline */
+  private fun updateCarePlanWithProtocol(carePlan: CarePlan, uris: List<CanonicalType>) {
```
pld, line 142: this doesn't appear to be used

android/engine/src/main/java/org/smartregister/fhircore/engine/task/WorkflowCarePlanGenerator.kt
```diff
@@ -0,0 +1,322 @@
...
+import org.hl7.fhir.r4.model.ResourceType
+import org.hl7.fhir.r4.model.Task
+
+@Singleton
+class WorkflowCarePlanGenerator
+@Inject
+constructor(
+  val knowledgeManager: KnowledgeManager,
+  val fhirOperator: FhirOperator,
+  val fhirEngine: FhirEngine,
```
pld, line 49: think we need to replace FhirEngine w/DefaultRepository and change all the calls to fhirEngine to call against defaultRepo

android/engine/src/main/java/org/smartregister/fhircore/engine/task/WorkflowCarePlanGenerator.kt
```diff
@@ -0,0 +1,322 @@
...
+        break
+      }
+    }
+  }
+
+  /**
+   * Find and update the status of the [CarePlan] activity as per the corresponding request resource
+   * status
+   */
+  suspend fun updateCarePlanActivity(
```
pld, line 256: this doesn't appear to be called

pld (changes requested):

android/engine/src/main/java/org/smartregister/fhircore/engine/task/WorkflowCarePlanGenerator.kt
owais-vd: calling from here https://github.com/opensrp/fhircore/blob/c65cdbbabdcd397bbe790dd0514b128a765fa459/android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirCarePlanGenerator.kt#L122-L127

android/engine/src/main/java/org/smartregister/fhircore/engine/task/WorkflowCarePlanGenerator.kt
owais-vd: removed

android/engine/src/main/java/org/smartregister/fhircore/engine/task/WorkflowCarePlanGenerator.kt
owais-vd: updated `fhirEngine` with `defaultRepository`

android/engine/src/main/java/org/smartregister/fhircore/engine/task/WorkflowCarePlanGenerator.kt
owais-vd: removed

android/engine/src/main/java/org/smartregister/fhircore/engine/task/WorkflowCarePlanGenerator.kt
pld: ah ok got it, so this is setting the status of each task-related activity in the careplan, not the status of the activity itself

pld (approved):

android/engine/src/main/java/org/smartregister/fhircore/engine/task/WorkflowCarePlanGenerator.kt
pld: ah missed that!

PR #2757, opened 2023-09-14
Practitioner supplychain report by practitioner

maimoonak:
Fixes [link to issue]

**Engineer Checklist**

**Code Reviewer Checklist**


pld (dismissed):
@AnnieMungai @dubdabasoduba not we're not prebuilding the practioner based reports so they will be slower to load, we can add that as a later feature, but as long as we're managing expections on practitioner based report build time I think this is ok as is for now

pld (commented):
@maimoonak this looks great, can you please add coverage for the new lines and take a look at this failing test, `MeasureReportRepositoryTest.testEvaluatePopulationMeasureHandlesBadMeasureUrl`

android/engine/src/main/java/org/smartregister/fhircore/engine/util/ReflectionUtil.kt
```diff
@@ -46,3 +48,22 @@ suspend fun <T : Any> KClass<T>.callSuspendFunctionOnField(
       .apply { isAccessible = true }
   return declaredObjectMethod.callSuspend(declaredObject, *args)
 }
+
+inline fun <reified T : Any> getPrivateProperty(property: String, obj: T): Any? {
```
ellykits, line 52: Was this implemented as workaround? Reflection code is less performant and would require code refactoring on API changes.

ellykits (changes requested):

pld:
@maimoonak is this a change to the upstream libraries that would be generally useful? If so, let's create an issue in the library that holds `org.opencds.cqf.cql.evaluator.measure.r4.R4MeasureProcessor` suggesting the changes we have here

maimoonak:
The PR had a fix for practitioner reports filter library parameter which has been merged in 0.2.5 but in main it can not be merged, the reason is main and this PR uses an old version of sdk-workflow which is rightnow incompatible with new version of sdk-engine.
We are using an old commit (595f422af2723b1993929965ad876172850d8f0c) to test and enhance reports on main rightnow and current main can not run measure reports due issues in sdk versions...
So effectively the PR is useless for main unless we have updated workflow library, and can be considered as blocked until we can make main stable with engine, data-capture, and workflow libs from latest sdk. The PR is not going to help with any working in main and should not be merged to avoid needless changes

pld:
Closing since we'll need to reimplement as @maimoonak noted

PR #2805, opened 2023-10-10
Affected Issue(s): Doc about "Setup Keycloak"

muhammad-levi:
What this commit has achieved:
1. Added documentation about setting up keycloak for OpenSRP v2.0


pld:
replaced with https://github.com/opensrp/fhircore/pull/2806

PR #2890, opened 2023-11-28
2849 allow option to hard code app id in properties and skip app id entry screen

### LINKED ISSUE #2849: Allow option to hard-code App ID in properties and skip App ID entry screen
Currently, the app workflow always requires the user to enter an App ID. Once this is entered it triggers a sync to the backend for all the apps configs & content.

Some of the apps will only ever expect an App ID, e.g. their icon and name might imply only a single App ID. This mean that we can hardcode the App ID in the APK and skip the App ID entry screen.

**Additional context**
- This should not modify any of the existing functionality when an App ID is not hard coded.
- Use the existing properties files where the FHIR API base URL is defined to define the App ID

**Acceptance criteria**
- [ ] when an App ID is defined in the built APK properties file
    - the App ID entry screen does not show
    - the App ID in the properties file is set anywhere an App ID entered in the screen would have been set
- [ ] if no App ID is defined, everything work as it currently does

**Area path**
1. Open an APK with the App ID in the properties file
    1. Verify the App ID entry screen does not show
1. Open an APK without the App ID in the properties file
    1. Verify the App ID entry screen does show

kelvin-ngure:

Fixes #2849

**Engineer Checklist**

**Code Reviewer Checklist**


android/properties.gradle.kts
```diff
@@ -41,3 +41,7 @@ val keystoreProperties = readProperties((project.properties["keystorePropertiesF
 requiredKeystoreProperties.forEach { property ->
   project.extra.set(property, keystoreProperties.getProperty(property,"sample_" + property))
 }
+
+// set APP_ID property
+val appIdProperty = "APP_ID"
```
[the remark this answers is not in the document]
kelvin-ngure: The appIdProperty variable holds the string that we will use as the key in the project.extra map. 

The flow is 
1. Extract APP_ID from localProperties using localProperties.get()
      if APP_ID was never added into the localProperties file as a value, it would return null so the """"""""" is the fallback. i.e if null set as empty string

2. After setting in project.extra, it is added as a buildConfigField  in the build.gradle section called defaultConfig{} that is used to autogenerate an object called BuildConfig at compile time
  
3. From there BuildConfig.APP_ID can be accessed from within the app. If we defined a value for APP_ID we get it, if we didn't we get an empty string as per the `line below` in your question

ellykits:
Can you link the issue to this PR? Why is this necessary?

kelvin-ngure:
> Can you link the issue to this PR? Why is this necessary?

https://github.com/opensrp/fhircore/issues/2849

android/properties.gradle.kts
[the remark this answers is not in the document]
ellykits: Convert `val appIdProperty = "APP_ID"` to a constant and declare it at the top of the file to improve the readability. 

```kotlin 
const val APP_ID_PROPERTY = "APP_ID"
```

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/appsetting/AppSettingActivity.kt
```diff
@@ -67,18 +70,35 @@ class AppSettingActivity : AppCompatActivity() {
     val existingAppId =
       sharedPreferencesHelper.read(SharedPreferenceKey.APP_ID.name, null)?.trimEnd()
 
+
+
     // If app exists load the configs otherwise fetch from the server
     if (!existingAppId.isNullOrEmpty()) {
```
ellykits, line 76: The only check we need to update will be this line. The function `loadConfigurations` will write the app id from the `local.properties` file to shared preferences.

```suggestion
    if (!existingAppId.isNullOrEmpty() || !BuildConfig.APP_ID.isNullOrEmpty()) {
```

The rest of the implementation need not to change.

ellykits (changes requested):

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/appsetting/AppSettingActivity.kt
ellykits: The `ConfigurationRegistry.loadConfigurations` will handle the error message if the application ID  is not found in the assets directory for debug variants.

kelvin-ngure: If we do that, the app would just freeze in the case of an non-debug app id e.g "wdf". This is because `ConfigurationRegistry.loadConfigurations()` would not be trying to load from assets and this piece of code in the same function does not throw anything when a composition does not exist i.e `?.` returns null


`else {
      fhirEngine.searchCompositionByIdentifier(parsedAppId)?.run {
        populateConfigurationsMap(context, this, false, parsedAppId, configsLoadedCallback)
      }
`

ellykits: I see. You are right, I missed the point that non-debug app configurations are downloaded from the server. My suggestion assumed the user had previously launched the app. Lemme re-check and revert.

ellykits: Borrowing from your implementation we can include the scenario when APP_ID Is provided in the local.properties in the if statement. Then use `fetchConfiguration` since it uses `loadConfiguration` internally. 

```kotlin
 // If app exists load the configs otherwise fetch from the server
    if (!existingAppId.isNullOrEmpty()) {
      appSettingViewModel.run {
        onApplicationIdChanged(existingAppId)
        loadConfigurations(appSettingActivity)
      }
    } else if (BuildConfig.OPENSRP_APP_ID.isNotEmpty()) {
      appSettingViewModel.run {
        onApplicationIdChanged(BuildConfig.OPENSRP_APP_ID)
        fetchConfigurations(appSettingActivity)
      }
    } else {
      setContent {
        AppTheme {
          val appId by appSettingViewModel.appId.observeAsState("")
          val showProgressBar by appSettingViewModel.showProgressBar.observeAsState(false)
          val error by appSettingViewModel.error.observeAsState("")
          AppSettingScreen(
            appId = appId,
            onAppIdChanged = appSettingViewModel::onApplicationIdChanged,
            fetchConfiguration = appSettingViewModel::fetchConfigurations,
            showProgressBar = showProgressBar,
            error = error,
          )
        }
      }
    }
```

The conditions should be in the order above. Should resolve the issue with app freezing.

ellykits: Can you also rename the config property to `OPENSRP_APP_ID`. There is an existing `APPLICATION_ID` configuration provided via Android gradle config, using `APP_ID` will be confusing.

android/quest/build.gradle.kts
```diff
@@ -65,6 +65,7 @@ android {
     buildConfigField("String", "OAUTH_BASE_URL", """"${project.extra["OAUTH_BASE_URL"]}"""")
     buildConfigField("String", "OAUTH_CLIENT_ID", """"${project.extra["OAUTH_CLIENT_ID"]}"""")
     buildConfigField("String", "OAUTH_SCOPE", """"${project.extra["OAUTH_SCOPE"]}"""")
+    buildConfigField("String", "APP_ID", """${project.extra["APP_ID"]}""")
```
ellykits, line 68:
```suggestion
    buildConfigField("String", "OPENSRP_APP_ID", """${project.extra["OPENSRP_APP_ID"]}""")
```

android/properties.gradle.kts
```diff
@@ -41,3 +41,7 @@ val keystoreProperties = readProperties((project.properties["keystorePropertiesF
...
+project.extra.set(appIdProperty, localProperties.getProperty(appIdProperty,  """""""" ))  // set empty string if property is null
```
ellykits, line 47: We can re-use the implementation for setting the required gradle properties by adding `OPENSRP_APP_ID` to this list `requiredFhirProperties`

ellykits: Then update the implementation here to default to empty String if `OPENSRP_APP_ID` is missing in the local.properties. Otherwise from the logic with the update it set the property to `sample_OPENSRP_APP_ID` which is not intended. Empty means no application ID was provided so we can proceed to display the settings page.

```kotlin
requiredFhirProperties.forEach { property ->
  project.extra.set(property, localProperties.getProperty(property, when {
    property.contains("URL") -> "https://sample.url/fhir/"
    property.equals("OPENSRP_APP_ID") -> ""
    else -> "sample_" + property
  }
  ))
}
```

ellykits (changes requested):

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/appsetting/AppSettingActivity.kt
ellykits: When using the OPENSRP_APP_ID from `local.properties`  we should not display the Setting screen at all to the user. It's okay for the application to force close with an error message if the configuration is missing.

android/properties.gradle.kts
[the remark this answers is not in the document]
ellykits: See my suggestion in the comment below, this property will no longer be necessary if we include `OPENSRP_APP_ID` to the list of `requiredFhirProperties`

ellykits (changes requested):

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/appsetting/AppSettingActivity.kt
kelvin-ngure: > When using the OPENSRP_APP_ID from `local.properties` we should not display the Setting screen at all to the user. It's okay for the application to force close with an error message if the configuration is missing.

Thank you for clearing this up!

ellykits (dismissed):

ellykits (approved):

PR #2931, opened 2023-12-20
sets color opacity on disable button

### LINKED ISSUE #2812: Set Color opacity on Compound Text & Button properties
Use case : Fade the buttons that the user should not be able to action 

Applies on
1. Compound Text
2. Actionable Button 
- Name of property   = `colorOpacity`
- To be applied to text color and background color

brandy-kay:

Fixes #2812

**Engineer Checklist**

**Code Reviewer Checklist**

### Demo
| Image of Enabled button                          | Image of Disabled button                                    |
| ----------------------------------- | ----------------------------------- |
| enabled_btn|diabled_btn|


android/quest/src/main/java/org/smartregister/fhircore/quest/ui/shared/components/ActionableButton.kt
```diff
@@ -99,7 +99,7 @@ fun ActionableButton(
               statusColor.copy(alpha = 0.08f)
             },
           contentColor = statusColor,
-          disabledBackgroundColor = DefaultColor.copy(alpha = 0.08f),
+          disabledBackgroundColor = statusColor.copy(alpha = 0.02f),
           disabledContentColor = DefaultColor,
```
ellykits, line 103: How about this disabled content color?

ellykits (changes requested):
The implementation here would work, is accepted and we can merge it. However, the issue was to implement a new configuration property for `CompoundText` and `ActionableButton` that will be used to change the opacity for the content and background colors used in those widgets. I think we can create a new issue for this PR labeled `Fix opacity color for disabled button`. I will add this comment to the issue tagged here.

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/shared/components/ActionableButton.kt
brandy-kay: The disabled content color had no effect on the button content so I opted not to interfere with it in the meantime.
Also if we are supposed to come up with the new config for the `CompoundText` and `ActionableButton` which of course is the right approach . I would consider we do just as you have suggested @ellykits

ellykits (approved):

PR #2934, opened 2023-12-20
Migrate SharedPreferences to DataStore

### LINKED ISSUE #2930: Migrate Shared preferences to datastore
Following this PR. We will migrate all the functionality of unencrypted shared preferences to datastore. 

**Additional context**
Scope
Replace function calls
Update view models with StateFlow when needed 
Write new tests
Remove unused test
Update documentation

**Acceptance criteria**
A check list of all things to verify once the implementation by the engineer is complete

**Area path**
1. Login to the app
2. Open _Navigation bar_
3. Click on _Children register_
4. Click on _Child profile_
5. Click on _Edit profile_ from menu

kelvin-ngure:

Fixes #2930

**Pull request highlights**

0. Moved Provision of the datastore to DataStore module. In test, Hilt injection overrides the dispatcher yet instantiation uses the correct testdispatcher. As a result, we don't use the DataStore module in test but rather ask the Faker to build a preferencesDataStore instance

1. Made a lot of the flows ready for observation from within the PreferencesDataStore.kt file. It therefore acts as a repository as well. These flows use the read() (now renamed to observe()) function from the previous PR. 

2. The read() function has been renamed to observe(). It is arguably a more apt description of what the flow does 

3. A new function called readOnce() has been introduced to allow us to capture the value of a flow when we do not want to subscribe to it’s emissions using the observe() function. Currently most, of the code in the engine needs this readOnce() behavior. Nevertheless, the observe() function will prove useful e.g in the feature to allow multiple users to log in on the same device or when consumed by StateFlows or LiveData in the UI.
This function is also useful for testing. It allows us to keep the tests unchanged in most of the places since it is a blocking read just like the sharedPreferences read (replacing reads with a flow would render all `verify{}` blocks useless since no function is called). Any need for structured concurrency can be addressed in the future for individual data stored. At this time, the focus is on introducing datastore into the codebase without refactoring too much to prevent adding bugs through what is likely to be a fairly large PR

5. Added a reified read() (and write()) that used gson for encoding similar to the one in the SharedPreferences. This is useful because 
	i. It is useful for storing List<String> structures and not having casts all over the code
       ii. It may be useful for temporarily storing structured data in the preferences datastore as we await decisions on what is to be stored in protodatastore 

6. location and locations have been renamed to locationIds and locationNames for readability. The same refactors have been applied to careTeam and careTeams. Similarly named keys have been created for them in the PreferencesDatastore

7. The observe() function Generic signature has changed from T  to T,M this allows the key and the output to have different types. This was necessary e.g when trying to observe a key value pair whose key is a string but whose reified datatype is a List<String> such as in the case of CARE_TEAM_IDS

8. Moved keys to once place. Using ResourceType.name ended up being restrictive. For example, we store, location ids and location names in SharedPreferences. ResourceType only has Location. This caused location ids to be stored under ResourceType.name (where ResourceType = Location) and location names be stored under SharedPreferenceKeys.location. As a result, any data being stored has a specific key name made for it within PreferencesDatastore.kt for code readability and flexibility of use. 

9. Will need discussion on whether my replacement of sharedPreferences in BaseMultiLanguageActivity is correct

**Engineer Checklist**

**Code Reviewer Checklist**


android/engine/src/main/java/org/smartregister/fhircore/engine/datastore/ProtoDataStore.kt
```diff
@@ -20,30 +20,30 @@ import android.content.Context
...
+          fileName = PRACTITIONER_DETAILS_DATASTORE_JSON,
+          serializer = PractitionerDetailsDataStoreSerializer,
   )
 
 val Context.userInfoProtoStore: DataStore<UserInfo> by
   dataStore(
-    fileName = USER_INFO_DATASTORE_JSON,
-    serializer = UserInfoDataStoreSerializer,
+          fileName = USER_INFO_DATASTORE_JSON,
+          serializer = UserInfoDataStoreSerializer,
```
ellykits, line 46: Looks like extra leading spaces were applied. Did you run `spotlessApply`?

kelvin-ngure: No, PR is not ready yet. Would you advise to do spotless on every commit?

ellykits: Okay. Running spotlessApply after every commit would be preferred. You can run it once when the PR is ready.

ellykits: Enable `ktfmt` to format the code directly in the IDE. Follow the setup here in the [Developer setup docs](https://docs.opensrp.io/engineering/android-app/developer-setup)

kelvin-ngure:
The most significant items left are 
1. Creating a Serializable wrapper for LocationHierarchies to allow storage
2. There's a single read I need to port that seems to use REMOTE_SYNC_RESOURCES
3. Updating tests that need proto datastore
4. writing any new tests and docs

ellykits:
Closing this PR in favor of https://github.com/opensrp/fhircore/pull/3205

PR #3011, opened 2024-01-24
Update application configuration documentation

sharon2719:

Fixes [link to issue]

**Engineer Checklist**

**Code Reviewer Checklist**

### DIFF
--- a/docs/engineering/android-app/configuring/config-types/application.mdx
+++ b/docs/engineering/android-app/configuring/config-types/application.mdx
@@ -7,7 +7,7 @@ title: Application
 These are app wide configurations used to control the application behaviour globally e.g. application theme, app language etc.
 
 :::note
-There can only be one instance of application configuration for the entire application. There are instances where the `Event Workflow` is added to the application config. See [here](https://docs.opensrp.io/engineering/android-app/configuring/event-management/resource-closure-by-background-worker) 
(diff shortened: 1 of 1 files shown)


docs/engineering/android-app/configuring/config-types/application.mdx
```diff
@@ -23,12 +23,39 @@ There can only be one instance of application configuration for the entire appli
     ],
     "useDarkTheme": false,
     "syncInterval": 15,
-    "syncStrategy": [
-        "Location",
-        "Organization",
-        "CareTeam",
-        "Practitioner"
+    "syncStrategy": "Location",
```
ellykits, line 26: @dubdabasoduba @ndegwamartin Confirm if this should be String or JsonArray? Do we support multiple sync strategies?

ndegwamartin: I don't think we should/do. For the case of the old OpenSRP where we updated the sync filter programatically to fetch depending on if supervisor - we now cater for that use case by using a different app-id for the different role where they can have their own strategy.

@dubdabasoduba can add on this

dubdabasoduba: I agree with @ndegwamartin. The combination of App Id and Sync Strategies caters for this.  

We might end up restructuring how we sync and tag resources but we can figure out the exact changes required when requirement gathering and scoping are done.

PR #3024, opened 2024-01-30
Rename tooling docs in side bar

pld:


pld:
this definitely feels like a flakey failure

```
FhirCarePlanGeneratorTest > generateOrUpdateCarePlan should generate careplan for 5 visits when lmp has passed 3 months FAILED
    org.junit.ComparisonFailure at FhirCarePlanGeneratorTest.kt:1030
```

@ndegwamartin can you please take a look?

Issue #3048, opened 2024-02-08
Add support for use of rules and computed values in Register configs and FAB


allan-on:
**Describe the feature request.**
Implement support for the use of rules and computed values in Register configs and FAB actions. 

**Additional context**
There's no support for rules in the register config (Other than `configRules`) outside of `registerCard` configs. 
Rules with access to the resource data can only be defined within the `registerCard`.  

**Acceptance criteria**
Rules based on the resource data can be defined outside the `registerCard`. 
Rules and the computed values should be accessible within the FAB actions. 

**Implementation plan (For Engineers)**
The plan for implementing the solution e.g. via a _description_ or a _check list_ for the various ordered tasks that will need to be completed.
_i.e. Describe how you intend to solve the problem_

allan-on:
This is no longer required because the configs were refactored here https://github.com/opensrp/fhircore/pull/3075
The configs introduced in that PR allow the use of the unique ID configs from register configs

PR #3075, opened 2024-02-16
Configure unique identifier assignment

ellykits:

Fixes [3048](https://github.com/opensrp/fhircore/issues/3048)

**Engineer Checklist**

**Code Reviewer Checklist**


ellykits:
This PR introduces a new configuration for unique ID assignments. Refer to the QuestionnaireConfig docs session to find out more. Here is a sample config:

```json
{
  "uniqueIdAssignment": {
    "linkId": "phn",
    "idFhirPathExpression": "Group.characteristic.where(exclude=false and code.text='phn').first().value.text",
    "readOnly": false,
    "resource": "Group",
    "sortConfigs": [
      {
        "paramName": "_lastUpdated",
        "dataType": "DATE",
        "order": "DESCENDING"
      }
    ],
    "resourceFilterExpression": {
      "conditionalFhirPathExpressions": [
        "Group.active = true and Group.type = 'device' and Group.name = 'Unique IDs'"
      ],
      "matchAll": true
    }
  }
}
```

docs/engineering/android-app/configuring/config-types/questionnaire.mdx
```diff
@@ -325,3 +326,43 @@ triggerConditions | This defines an array of condition for to be met for the eve
 eventResourceId |  uniqueId of resource id to be closed |  yes   |               |
 eventResources | A list of resources to close(Type of ResourceConfig) |    yes           |     |      |
 
+
```
dubdabasoduba, line 329: @ellykits  do we dictate where in the Patient, Group e.t.c resource is added?

ellykits: Which resources are you referring to here @dubdabasoduba? The resource used to sync unique identifiers? This new configuration (see the sample in the comment) is added to the Questionnaire config.

docs/engineering/android-app/configuring/config-types/questionnaire.mdx
```diff
@@ -325,3 +326,43 @@ triggerConditions | This defines an array of condition for to be met for the eve
...
+## Unique ID assignment
+
```
pld, line 331: @allan-on can you please add a sub-section here describing

pld (dismissed):

docs/engineering/android-app/configuring/config-types/questionnaire.mdx
```diff
@@ -325,3 +326,43 @@ triggerConditions | This defines an array of condition for to be met for the eve
...
+Here is as sample configuration for the unique identifier assignment:
+
+```json
+{
+  "uniqueIdAssignment": {
+    "linkId": "phn",
+    "idFhirPathExpression": "Group.characteristic.where(exclude=false and code.text='phn').first().value.text",
+    "readOnly": false,
+    "resource": "Group",
```
pld, line 343: let's make that for now this only works for the `Group` resource

pld:
@allan-on please add a top level feature for unique ID assignment here, https://docs.opensrp.io/features/app-features, then tag @rowo and me to review it

ellykits:
@allan-on I will add the remaining tests to improve the coverage during CHT this week.

pld:
> @allan-on I will add the remaining tests to improve the coverage during CHT this week.

can add an issue that refs this pr?

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireViewModel.kt
```diff
@@ -274,13 +304,37 @@ constructor(
...
       onSuccessfulSubmission(idTypes, currentQuestionnaireResponse)
     }
   }
 
+  suspend fun retireUsedQuestionnaireUniqueId() {
+    if (uniqueIdResourcePair != null) {
+      val (id, resource) = uniqueIdResourcePair!!
+
+      // Update Group resource. Can be extended in future to support other resources
+      if (resource is Group) {
```
allan-on, line 321: @ellykits In this implementation, prepoulated IDs are assumed to be used and therefore marked as used on Questionnaire submission. Since it is possible to edit the ID in the questionnaire (if not read-only), it is possible that the ID submitted with the questionnaire isn't the same as the one that was originally prepopulated. 
We might have to confirm by checking from the `QuestionnaireResponse`

ellykits: Is this because the field is editable?

pld: Interesting, good catch. I agree, it makes sense to check the QR and only mark it as used if the ID in the QR matches the one we are expecting to be used.

It seems OK for whether it's editable or not to be an implementation decisions, as long as the implementer understands that is it is editable, the IDs can no longer be guaranteed unique, let's note this in the docs

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireViewModel.kt
```diff
@@ -186,6 +190,32 @@ constructor(
...
+                questionnaireConfig.uniqueIdAssignment,
+              )
+
+            val extractedId =
+              fhirPathDataExtractor.extractValue(
+                base = resource,
+                expression = uniqueIdAssignmentConfig.idFhirPathExpression,
+              )
+            if (resource != null && extractedId.isNotEmpty()) {
+              uniqueIdResourcePair = Pair(extractedId, resource) // To be updated upon submission
```
pld, line 209: since we're looking at the QR, do we need to set the state here? I'd simply not set the state if we do not have to, if we do need this in `retireUsedQuestionnaireUniqueId`, can we pass it as an argument?

docs/engineering/android-app/configuring/config-types/questionnaire.mdx
```diff
@@ -325,3 +326,43 @@ triggerConditions | This defines an array of condition for to be met for the eve
...
+Unique ID configs are used to configure how generated unique ID is extracted from a FHIR resource. Currently the Group Resource
+is used to sync generated identifiers. 
+
+Here is as sample configuration for the unique identifier assignment:
+
+```json
+{
+  "uniqueIdAssignment": {
+    "linkId": "phn",
+    "idFhirPathExpression": "Group.characteristic.where(exclude=false and code.text='phn').first().value.text",
```
pld, line 341: Can you explain why this should be configurable? I think you did a great job writing the code to work based on how this is configured, but that leads to some performance compromises

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireViewModel.kt
```diff
@@ -282,13 +312,43 @@ constructor(
...
+      if (resource is Group) {
+        resource.characteristic.onEach {
+          if (it.hasValueCodeableConcept() && it.valueCodeableConcept.text == submittedUniqueId) {
+            it.exclude = true
+            return@onEach
+          }
+        }
+
+        // Mark Group as inactive when all IDs are retired
+        resource.active = resource.characteristic.any { !it.exclude }
```
pld, line 344: I'm worried about the performance implications of doing 2 full scans of the characteristic list.

If we do not allow the ID FHIRPath expression to be configurable, we can use a counter to do this and have everything in constant time. We can set `group.quantity` to the index of the next unused ID, then all we do here is say `resource.quantity++`. Then set to readonly if `resource.quantity > resource.characteristic.size`

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireViewModel.kt
ellykits: Yes, this state is not necessary. The ID can be obtained from the QuestionnaireResponse.

docs/engineering/android-app/configuring/config-types/questionnaire.mdx
ellykits: 1. The first assumption was we'd not expect multiple items in Group#characteristics
2. This was configured to support other ResourceTypes in the future; not to limit the functionality to Group resources. Extracting data from resources using FHIR path was also an easier alternaitve compared to hard-coding implementation.

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireViewModel.kt
ellykits: The recommended approach will work if we hard-code this functionality for a Group resource. However, the [Group.quantity](http://hl7.org/fhir/R4B/group-definitions.html#Group.quantity) property is supposed to be used as the count of the members according to the FHIR documentation.

ellykits: A boolean state can be updated in the first iteration of the characteristics to be used as the value for `resource.active` to prevent the second scan.

pld: Yea true and true, but we only support group now so we don't lose anything by making this only work for group, since we're already not following the semantics of characteristic, seems fine to not follow that for quantity

pld: But this would be a meaningful change, late in the process, for efficiency reasons I'm ok w/keeping this as is if we've run some tests with a Group of reasonable size on low end devices and this has negligible/no impact on the performance

ellykits: @allan-on Can you verify the performance impact? Do you know on average how many items we would expect in the `Group.characteristics` property?

pld (approved):

pld:
Looks like `testRetrieveUniqueIdAssignmentResourceShouldReturnAResource` is failing there's more coverage needed for `retrieveUniqueIdAssignmentResource`, along with a couple other functions

pld (approved):

Issue #3077, opened 2024-02-16
[New Feature] Add Related Entity Location Tag to newly created resources


f-odhiambo:
**Background & Context**

As a health practitioner, I want to be able to distinguish patients based on the location they are located at.

As a data collector or service monitor, I want to be able to distinguish information related to a point of interest based on its connection to that point of interest.

Also see discussion item 
- https://docs.google.com/document/d/1D7fXYo3zFMAO3j5Vh8FmpfSVeRfxfVjHAefDtw2KmSk/edit
- https://github.com/ona-health/sid-dto-fhir-resources/discussions/753 

**Proposed Solution**

High level:
- Add a `meta.tag` called `Related Entity Location` to every resource generated on the app that is tied to a *related entity* that could be a Patient, a Group, or a Location (eg representing a service point or POI).
- The *related entity* is itself or is connected to a location at its time of creation.

Assigning a value to `meta.tag: "Related Entity Location"` can happen in 3 ways

1. Using a location ID submitted in a Questionnaire and stored in a QR (QuestionnaireResponse). 
    - This will happen during entity creation, e.g. creation of a Patient, Group (e.g. representing a family), or Location (e.g. represent a service point)
    - When creating the entity the users selects a location in the registration questionnaire, this is saved in the QR
    - In extraction a StructureMap creates a Location resource, or references an existing Location
    - This location is stored in the `meta.tag: "Related Entity Location"` of the *related entity* (in this case the *related entity* is the *entity* itself).
2. When saving a resource connected to a related entity, set to the value in the related entity
    - This applies when the registration above has already happened and the data is generated through other operations, e.g the generation of CarePlans and Tasks via $apply or the extraction of Observations via StructureMaps.
    - We assume a `subject` or similar field links the newly saved resource to the *related entity*
    - Set `[new resource].meta.tag: "Related Entity Location"` = `[related entity].meta.tag: "Related Entity Location"`
 
Example meta tag:
````json
{
  "system": "https://smartregister.org/related-entity-location-tag-id",
  "code": "1a8ae4f2-c94b-4366-ac31-f55ccf3168a2",
  "display": "Related Entity Location"
}
````

Patient resource with meta-tag:
````json
{
  "resourceType": "Patient",
  "id": "21248",
  "meta": {
    "versionId": "1",
    "lastUpdated": "2023-06-20T12:43:56.633+00:00",
    "source": "#e58f8eb4117b05a8",
    "tag": [
      {
[...]

PR #3086, opened 2024-02-20
Add Related entity location tag on newly created resources via configs

ellykits:

Fixes https://github.com/opensrp/fhircore/issues/3077

**Engineer Checklist**

**Code Reviewer Checklist**

### DIFF
--- a/android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirCarePlanGenerator.kt
+++ b/android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirCarePlanGenerator.kt
@@ -16,6 +16,7 @@
 
 package org.smartregister.fhircore.engine.task
 
+import android.content.Context
 import androidx.annotation.VisibleForTesting
 import ca.uhn.fhir.context.FhirContext
 import ca.uhn.fhir.util.TerserUtil
@@ -23,6 +24,7 @@ import com.google.android.fhir.FhirEngine
 import com.google.android.fhir.get
 import com.google.android.fhir.logicalId
 import com.google.android.fhir.search.search
+import dagger.hilt.android.qualifiers.ApplicationContext
 import java.util.Date
 import javax.inject.Inject
 import javax.inject.Singleton
@@ -33,6 +35,7 @@ import org.hl7.fhir.r4.model.Bundle
 import org.hl7.fhir.r4.model.CanonicalType
 import org.hl7.fhir.r4.model.CarePlan
 import org.hl7.fhir.r4.model.CodeableConcept
+import org.hl7.fhir.r4.model.Coding
 import org.hl7.fhir.r4.model.DateTimeType
 import org.hl7.fhir.r4.model.Dosage
 import org.hl7.fhir.r4.model.Expression
@@ -51,6 +54,7 @@ import org.hl7.fhir.r4.model.Timing
 import org.hl7.fhir.r4.model.Timing.UnitsOfTime
 import org.hl7.fhir.r4.utils.FHIRPathEngine
 import org.hl7.fhir.r4.utils.StructureMapUtilities
+import org.smartregister.fhircore.engine.R
 import org.smartregister.fhircore.engine.configuration.QuestionnaireConfig
 import org.smartregister.fhircore.engine.configuration.event.EventType
 import org.smartregister.fhircore.engine.data.local.DefaultRepository
@@ -76,6 +80,7 @@ constructor(
   val defaultRepository: DefaultRepository,
   val fhirResourceUtil: FhirResourceUtil,
   val workflowCarePlanGenerator: WorkflowCarePlanGenerator,
+  @ApplicationContext val context: Context,
 ) {
   private val structureMapUtilities by lazy {
     StructureMapUtilities(transformSupportServices.simpleWorkerContext, transformSupportServices)
@@ -99,6 +104,11 @@ constructor(
     data: Bundle = Bundle(),
     generateCarePlanWithWorkflowApi: Boolean = false,
   ): CarePlan? {
+    val relatedEntityLocationTags =
+      subject.meta.tag.filter {
+        it.system == context.getString(R.string.sync_strategy_related_entity_location_system)
+      }
+
     // Only one CarePlan per plan, update or init a new one if not exists
     val output =
       fhirEngine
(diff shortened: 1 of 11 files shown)


android/engine/src/main/java/org/smartregister/fhircore/engine/util/extension/ResourceExtension.kt
```diff
@@ -300,6 +303,40 @@ fun Resource.appendPractitionerInfo(practitionerId: String?) {
...
+          .find(linkIdConfig.linkId)
+          ?.itemFirstRep
+          ?.answerFirstRep
+          ?.valueStringType
+          ?.value
+
+      val existingTag = this.meta.getTag(locationCoding.system, locationId)
+      if (!locationId.isNullOrEmpty() && existingTag == null) {
+        this.meta.addTag(locationCoding.apply { setCode(locationId) })
+      }
```
pld, line 336: this seems like the cheaper operation (is that correct?), I'd move it above the locationID and only get locationID if this as null, don't even need to assign this, eg s/t like
```suggestion
      if(this.meta.getTag(locationCoding.system, locationId) == null) {
          val locationId =
            questionnaireResponse
              .find(linkIdConfig.linkId)
              ?.itemFirstRep
              ?.answerFirstRep
              ?.valueStringType
              ?.value
    
          if (!locationId.isNullOrEmpty()) {
            this.meta.addTag(locationCoding.apply { setCode(locationId) })
      }
```

ellykits: Retrieving the `locationId` can be expensive depending on the number of items to be iterated. This can be done outside this function and the id passed as a parameter.

pld: I see, makes sense, cool

android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirCarePlanGenerator.kt
```diff
@@ -99,6 +104,11 @@ constructor(
     data: Bundle = Bundle(),
     generateCarePlanWithWorkflowApi: Boolean = false,
   ): CarePlan? {
+    val relatedEntityLocationTags =
+      subject.meta.tag.filter {
+        it.system == context.getString(R.string.sync_strategy_related_entity_location_system)
```
pld, line 109: I would not expect `system` to be unique, wouldn't it be code that is the unique item to filter on? Isn't [meta.tag a coding](https://hl7.org/fhir/R4B/resource.html#Meta), or is this the wrong reference?

pld (changes requested):

android/engine/src/main/java/org/smartregister/fhircore/engine/task/FhirCarePlanGenerator.kt
ellykits: The idea here is we get all the tags using the system for `RelatedEntityLocation` coding. In the current use case, this should return a single tag.

ellykits: At this point, propagating the locationId that's used as the code is not so much useful

pld: The issue is that `system` means the terminology system, like ICD11 or something, even though we're not using it that way, so this wouldn't return the correct set of tags if we used it correctly.

So I guess I'm suggesting 2 changes (that contradict what's written in the issue, apologies that I did not flag this earlier)

1. We use system correctly, so change this to "https://smartregister.org/" if we say we are using our own internal terminology system
2. define and store in docs a "code" that means "Related Entity Location" and filter by "code" here

What do you think?

cc @dubdabasoduba

ellykits: I see. This means we'd have to update the other tags as well with this interpretation of the system.

ellykits: Sample of all meta tags.

```json
"tag": [
      {
        "system": "urn:oid:2.16.578.1.12.4.1.1.8655",
        "display": "Patient Registration",
        "_display": {
          "extension": [
            {
              "url": "http://hl7.org/fhir/StructureDefinition/translation",
              "extension": [
                {
                  "url": "lang",
                  "valueCode": "id"
                },
                {
                  "url": "content",
                  "valueString": "Pendaftaran Pasien"
                }
              ]
            }
          ]
        }
      },
      {
        "system": "https://smartregister.org/related-entity-location-tag-id",
        "code": "33f45e09-f96e-41d3-9916-fb96455a4cb2",
        "display": "Related Entity Location"
      },
      {
        "system": "https://smartregister.org/care-team-tag-id",
        "code": "47d68cac-306f-4b75-9704-b4ed48b24f76",
        "display": "Practitioner CareTeam"
      },
      {
        "system": "https://smartregister.org/location-tag-id",
        "code": "Not defined",
        "display": "Practitioner Location"
      },
      {
        "system": "https://smartregister.org/organisation-tag-id",
        "code": "ca7d3362-8048-4fa0-8fdd-6da33423cc6b",
        "display": "Practitioner Organization"
      },
      {
        "system": "https://smartregister.org/practitioner-tag-id",
        "code": "9db48504-9f63-411b-b61e-28351d7af5e8",
        "display": "Practitioner"
      },
      {
        "system": "https://smartregister.org/app-version",
        "code": "1.1.0-sidBunda",
        "display": "Application Version"
      }
    ]
```

pld: Yea I'm just knowing that, I don't want to do that all at once do, we can make sure we do it correctly here and then change the remainder over time. I'm checking in w/ @dubdabasoduba and @f-odhiambo to confirm we're all on the same page...

ellykits: For uniqueness, we've been using a combination of system + code

pld: @dubdabasoduba raised good points on why we do it this way and that there's a perspective through which it is semantic, I will write up some docs based on the conversation and resolve this conversation, we can move ahead as is, thank you for the details

pld (dismissed):

pld (approved):

PR #3097, opened 2024-02-23
Use subject's Related Entity Location meta tag on related resources

ellykits:

Fixes https://github.com/opensrp/fhircore/issues/3077

**Engineer Checklist**

**Code Reviewer Checklist**

### DIFF
--- a/android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireViewModel.kt
+++ b/android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireViewModel.kt
@@ -430,14 +430,13 @@ constructor(
       }
     if (resourceIdPair != null) {
       val (resourceType, resourceId) = resourceIdPair
-      val group =
-        loadResource(resourceType = resourceType, resourceIdentifier = resourceId) as Group?
-      if (group != null) {
+      val resource = loadResource(resourceType = resourceType, resourceIdentifier = resourceId)
+      if (resource != null) {
         val system =
           context.getString(
             org.smartregister.fhircore.engine.R.string.sync_strategy_related_entity_location_system,
           )
-        group.meta.tag.filter { coding -> coding.system == system }.forEach(this.meta::addTag)
+        resource.meta.tag.filter { coding -> coding.system == system }.forEach(this.meta::addTag)
       }
     }
   }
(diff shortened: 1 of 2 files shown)


pld (approved):

pld (dismissed):

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireViewModel.kt
```diff
@@ -411,23 +414,30 @@ constructor(
...
+      val (resourceType, resourceId) = resourceIdPair
+      val group =
+        loadResource(resourceType = resourceType, resourceIdentifier = resourceId) as Group?
+      if (group != null) {
+        val system =
+          context.getString(
+            org.smartregister.fhircore.engine.R.string.sync_strategy_related_entity_location_system,
+          )
+        group.meta.tag.filter { coding -> coding.system == system }.forEach(this.meta::addTag)
       }
```
dubdabasoduba, line 441: @ellykits  how do we handle the situation where we do not want to propagate the Group REL (Related Entity location) to the patient resources?

dubdabasoduba (changes requested):

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/questionnaire/QuestionnaireViewModel.kt
ellykits: The Group's REL only propagates if the Questionnaire has a `groupResource` (should only happen when adding Group member) config. If we do not want to propagate the group's REL two options:

1. Do not configure `groupResource` on resource 
2. Provide configurations for extracting location ID from QuestionnaireResponse when launching the questionnaire. The location ID will be used on the subject and related resources extracted from the questionnaire submission. (To be used when you want the configured locationId to override the subject's)

The subject resource's REL is given precedence over the Group's REL because, if the subject were added as a member of the Group it would have the Group's REL anyway.

dubdabasoduba: @ellykits  I assume point 2 is going to handle the situation where a Familly member does not leave in the same place as the rest of the family, right ?

ellykits: Yes that's correct.

f-odhiambo (approved):
Resource are adding Tags as expected

dubdabasoduba (approved):

PR #3101, opened 2024-02-26
Move custom .kts to buildSrc as precompiled plugins

LZRS:
https://docs.gradle.org/8.2/userguide/custom_plugins.html#sec:precompiled_plugins

Fixes [link to issue]

**Engineer Checklist**

**Code Reviewer Checklist**


android/buildSrc/src/main/kotlin/project-properties.gradle.kts
```diff
@@ -1,17 +1,19 @@
...
   val properties = Properties()
   val localProperties = File(file)
   if (localProperties.isFile) {
     InputStreamReader(FileInputStream(localProperties), Charsets.UTF_8).use { reader
       ->
       properties.load(reader)
     }
-  } else if(file.toString().contains(File.separator)) throw FileNotFoundException("\u001B[32mFile $file not found\u001B[0m")
+  }
+//  else if(file.toString().contains(File.separator)) throw FileNotFoundException("\u001B[32mFile $file not found\u001B[0m")
```
ndegwamartin, line 16: Should we remove this then?

android/buildSrc/src/main/kotlin/project-properties.gradle.kts
```diff
@@ -1,17 +1,19 @@
...
   else  println("\u001B[34mFILE_NOT_FOUND_EXCEPTION: File $file not found\u001B[0m")
```
ndegwamartin, line 17: For the else condition here, perhaps we should throw the exception then instead of printing an error because if someone specifies a file to be read they expect that it will be found. If it isn't that is most likely an error and the APK will be built with the default sample credentials _(which will make the app fail during runtime)_

PR #3125, opened 2024-03-05
Add pagination to list configs end point during config sync

### LINKED ISSUE #3049: Add support for paginated List configurations endpoint during Config sync
Following the optimisation [refactor to the plugin gateway here](https://github.com/onaio/fhir-gateway-plugin/pull/32), we need to refactor the client to take advantage of the pagination.

**Additional context**
Currently, when list entries are configure in the composition the whole list of references is fetched in one batch. This can lead to a high latency when processing the configs especially when we have a large number of entries. In some cases the connection will time out.
 
**Acceptance criteria**

- [ ] Refactor the ConfigRegistry class post/processing of List entries in the Composition to utilize pagination
- [ ] Configurations sync should work correctly
- [ ] Unit tests

**Area path**
1. Login to the app - e.g. SID app which uses the location hierarchy
2. Perform sync to completion
3. Go to **Household** and **Add member**. Use the location drop downs to assigning user location
4. Save the record and confirm extraction was successful

In the Config Registry class, add _page, _count and the List resource type in the function that fetches List Resources. 
Check the logs for the result of adding pagination.
Refactor ConfigRegistry tests.

Raynafs:

Fixes #3049

**Engineer Checklist**

**Code Reviewer Checklist**


Raynafs:
Logs showing List Resource being fetched after adding pagination.

Raynafs:
image
image (2)

android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/ConfigurationRegistry.kt
```diff
@@ -696,7 +696,8 @@ constructor(
                       TYPE_REFERENCE_DELIMITER,
                     )
                   val resourceId = listEntryComponent.item.reference.extractLogicalIdUuid()
-                  val listResourceUrlPath = "$resourceKey?$ID=$resourceId&_count=$DEFAULT_COUNT"
+                  val listResourceUrlPath =
+                    "${ResourceType.List.name}?$resourceKey?$ID=$resourceId&_page&_count"
```
SebaMutuku, line 700: @Raynafs we might have to add a default page and count if someone hasn't provided them

SebaMutuku (changes requested):

android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/ConfigurationRegistry.kt
Raynafs: Oh yes sure. Adding it

android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/app/ApplicationConfiguration.kt
```diff
@@ -32,6 +32,7 @@ data class ApplicationConfiguration(
   val syncInterval: Long = 15,
   val syncStrategies: List<String> = listOf(),
   val loginConfig: LoginConfig = LoginConfig(),
+  val listSyncConfig: ListSyncConfig = ListSyncConfig(),
```
ndegwamartin, line 35: For new sync configurations we also need to add documentation for the docusite here https://docs.opensrp.io/engineering/android-app/configuring

Raynafs: Oh yes sure @ndegwamartin I will do so

android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/ConfigurationRegistry.kt
```diff
@@ -696,7 +701,8 @@ constructor(
                       TYPE_REFERENCE_DELIMITER,
                     )
                   val resourceId = listEntryComponent.item.reference.extractLogicalIdUuid()
-                  val listResourceUrlPath = "$resourceKey?$ID=$resourceId&_count=$DEFAULT_COUNT"
+                  val listResourceUrlPath =
+                    "${ResourceType.List.name}?$resourceKey?$ID=$resourceId&_page=${applicationConfig.listSyncConfig.page} &_count= ${applicationConfig.listSyncConfig.count}"
```
ndegwamartin, line 705: This appears to be adding functionality to the Non-Proxy code block. This Pagination feature is implemented on the OpenSRP FHIR Gateway https://github.com/onaio/fhir-gateway-extension/issues/29

Also see the corresponding PR https://github.com/onaio/fhir-gateway-extension/pull/32

Raynafs: Oooh yeah. Correcting that

android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/app/ApplicationConfiguration.kt
Raynafs: Yes sure, I set the page number to 1 in the app_config.json

android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/app/ListSyncConfig.kt
```diff
@@ -0,0 +1,25 @@
...
+
+package org.smartregister.fhircore.engine.configuration.app
+
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class ListSyncConfig(
+  val page: Int = 1,
+  val count: Int = 20,
+)
```
ellykits, line 25: This new configuration can be hardcoded to a reasonable number unless there is a use case for the flexibility.

Raynafs: @ellykits So I put it that way because those are the default values on the server side. What would be the recommended value for us on the client side?

Raynafs:
Image of debug mode showing the fetched  listResourceUrlPath
image

android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/app/ListSyncConfig.kt
ndegwamartin: We can hardcode to 50 and test out how it behaves.In the end it will depend on the backend performance by we can start with this and maybe tweak upwards or downwards. 

Also based on my [previous comment here](https://github.com/opensrp/fhircore/pull/3125#discussion_r1522715292), we don't actually need to make pageNumber configurable. It can be hardcoded as well for the first fetch

android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/ConfigurationRegistry.kt
```diff
@@ -714,6 +730,13 @@ constructor(
           searchPath = "${resourceGroup.key}/${it.focus.extractId()}",
           patientRelatedResourceTypes = patientRelatedResourceTypes,
         )
+        if (resourceGroup.key.contains("List")) {
+          val searchPath = "${resourceGroup.key}/${it.focus.extractId()}"
+          val resourceId = searchPath.substringAfter(TYPE_REFERENCE_DELIMITER).trim()
+          val listResourceUrlPath =
+            "${ResourceType.List.name}?$ID=$resourceId&_page=${applicationConfig.listSyncConfig.page}&_count=${applicationConfig.listSyncConfig.count}"
+          fhirResourceDataSource.getResource(listResourceUrlPath)
+        }
```
ndegwamartin, line 739: I don't think we need this check and code block here. The parent method `processCompositionListResources` solely processes List resource type

android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/app/ListSyncConfig.kt
Raynafs: Oooh okay. So Lemme change the number to 50. Then I also hard code the pageNumber and page count as well. @ndegwamartin should I then remove the listSync from the app config completely?

android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/ConfigurationRegistry.kt
Raynafs: Ooh okay. Changing that as well

ndegwamartin (changes requested):
This implementation will fetch the first batch/page successfully, however we still need the processing of subsequent pagination pages implemented.

Raynafs:
Oh okay. Lemme add the functionality

Raynafs:
image
This specific list has a total of 38 items hence since we have put a count of 200. Page one will hold all the 38 items.

Raynafs:
image
Changed the count to 4

android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/ConfigurationRegistry.kt
```diff
@@ -491,24 +491,56 @@ constructor(
...
-      } else
-        fhirResourceDataSource.getResourceWithGatewayModeHeader(
-          gatewayModeHeaderValue,
-          searchPath,
-        )
+    val pageSize = 200
+    var nextPage = currentPage
+    var totalItemsProcessed = 0
+
+    while (true) {
```
ndegwamartin, line 500: I don't think we need this here, or rather we should refactor to remove it and use a different approach.

android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/ConfigurationRegistry.kt
```diff
@@ -491,24 +491,56 @@ constructor(
...
+      val resultBundle = fetchResourceBundle(gatewayModeHeaderValue, searchPath, nextPage)
+      val itemsCount = resultBundle.entry.size
+      totalItemsProcessed += itemsCount
+
+      processResultBundleEntries(resultBundle.entry, patientRelatedResourceTypes)
+
+      when {
+        itemsCount == pageSize -> {
```
ndegwamartin, line 508: This condition seems to terminate the loop after the hardcoded pagesize of 200 is reached, would the solution still work if we had more than 200 pages?

Raynafs:
[Uploading Screencast from 03-04-2024 12:46:53 ALASIRI.webm…]()

Raynafs:
[Screencast from 03-04-2024 12:46:53 ALASIRI.webm](https://github.com/opensrp/fhircore/assets/110402503/8b5a0d2b-2199-4699-bbaa-7c7074799fd7)
@ndegwamartin

android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/ConfigurationRegistry.kt
```diff
@@ -491,24 +491,52 @@ constructor(
...
   ) {
-    val resultBundle =
-      if (gatewayModeHeaderValue.isNullOrEmpty()) {
-        fhirResourceDataSource.getResource(searchPath)
-      } else
-        fhirResourceDataSource.getResourceWithGatewayModeHeader(
-          gatewayModeHeaderValue,
-          searchPath,
-        )
+    var currentPageUrl = "$searchPath&_page=$currentPage&_count=4"
```
ndegwamartin, line 496: we can remove the hard coded 4 to pick from the constants. during dev test mode you can change the constant value instead

Raynafs: Oh okay. Lemme correct that

android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/ConfigurationRegistry.kt
```diff
@@ -491,24 +491,52 @@ constructor(
...
+
+
+    // Fetch pages until there are no more next pages
+    while (currentPageUrl.isNotBlank()) {
```
ndegwamartin, line 500: Instead of using a `while` loop, why don't we just read the response from **resultBundle** and then make the next request if it exists?

android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/ConfigurationRegistry.kt
```diff
@@ -491,24 +491,52 @@ constructor(
...
+        break
+      }
 
-    processResultBundleEntries(resultBundle, patientRelatedResourceTypes)
+      // Extracting page number from URL to increment it for the next page
+      val regex = Regex(".*&_page=(\\d+).*")
+      val matchResult = regex.matchEntire(currentPageUrl)
+      val nextPage = matchResult?.groupValues?.get(1)?.toIntOrNull()
+      nextPage?.let {
+        currentPageUrl = currentPageUrl.replace("_page=$nextPage", "_page=${nextPage + 1}")
```
ndegwamartin, line 519: We already have the full url required for the next fetch so no need to parse the contents and extract the params. This way we can avoid a performance hit

android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/ConfigurationRegistry.kt
Raynafs: The while loop helps me to fetch all the pages. The latter implementation only allows me to fetch the first page (instead of fetching all the total number of items). So even when I put the if (resultBundle.link("next")) it only returns the first page that was fetched. I am not sure if there is a url that can assist to fetch all the items in the gateway.

ndegwamartin: I think `next` should always have the next page request

android/quest/src/test/java/org/smartregister/fhircore/quest/app/ConfigurationRegistryTest.kt
```diff
@@ -177,22 +177,32 @@ class ConfigurationRegistryTest : RobolectricTest() {
         addEntry().apply {
           this.resource = ListResource().apply { ListResource@ this.id = "123456" }
         }
+        link.add(
+          Bundle.BundleLinkComponent().apply { relation = "next" },
```
ndegwamartin, line 181: `PAGINATION_NEXT` constant

ndegwamartin (approved):

PR #3259, opened 2024-05-14
QR to HTML Population

### LINKED ISSUE #3257: Populate HTML Content with QuestionnaireResponse Data
Create a feature that's able to:
- Hide UI elements if there is no answer in the referred link id
- Populate answers as a list or a single value in the HTML
- Populate the submitted date in the HTML content

**Additional context**
I was initially thinking to use configuration + rules engine, but I feel it will take much more time since the implementation of those 2 need to be refactored first before it can be used by this feature. So, that's why I create a new class then do some simple string manipulation to the html.

**Acceptance criteria**
- [x] Hide UI elements if there is no answer in the referred link id
 ```
If the answer is empty:
"@is-not-empty('link-id')<p>Text</p>@is-not-empty" -> (nothing)
     
If the answer is not empty:
"@is-not-empty('link-id')<p>Text</p>@is-not-empty" -> <p>Text</p>
```

- [x] Populate answers as a list or a single value in the HTML

```
"</ul>@answer-as-list('link-id')</ul>" -> <ul><li>answer 1</li><li>answer 2</li></ul>
```

```
"<p>@answer('link-id')</p>" -> <p>(the answer)</p>
     
"<p>@answer('link-id')</p>" -> <p>14-May-2024</p> (if it's a date type)
     
"<p>@answer('link-id','MMMM d, yyyy')</p>" -> <p>May 14, 2024</p>
```

- [x] Populate the submitted date in the HTML content
```
"@submitted-date" -> 14-May-2024

"@submitted-date('MMMM d, yyyy')" -> May 14, 2024
```

**Area path**
Run the test cases, we couldn't directly use this feature in the app until ticket https://github.com/opensrp/fhircore/issues/3235 is complete.

1. Create HtmlPopulator class
2. Create populateHtml() method, it will accept a raw HTML string and return the populated HTML string with the data from the QuestionnaireResponse
3. Create these methods: processIsNotEmpty(), populateAnswerAsList(), populateAnswer(), populateSubmittedDate(). These will handle specific tasks such as hiding UI elements, populating answers as list items or single values, and populating the submitted date in the HTML content, mainly using string manipulation by finding the tags for each functionality, then replacing it with the correct string/answer.

FikriMilano:

Fixes #3257

**Engineer Checklist**

**Code Reviewer Checklist**


ellykits (dismissed):

ndegwamartin:
This implementation could be refactored for some optimization e.g. use one while loop with if else statements for the various conditions during parsing or using regex for the various parse/replace logic.

The performance gains are probably negligible for the size of HTML we are parsing so will go ahead and approve this. We can keep a look out and perform any refactors accordingly incase latency is noticable.

ndegwamartin:
cc @FikriMilano

ndegwamartin (dismissed):

android/engine/src/main/java/org/smartregister/fhircore/engine/pdf/HtmlPopulator.kt
pld, line 1: agree with @ndegwamartin, let's refactor this now. I think a while loop would make sense if the substitutions we are planning may contain data that will then again need to substituted, but that does not appear to be expected here.

As @ndegwamartin suggested, let's replace the 3 while loops with a single pass over the data to perform the substitutions. I also do not think a while is what we want, because we have to do a search over the data the number of replacements we are making, i.e. suppose the length of the data is N and we are making M replacements, this runs in O(NM), we have to search the data M times, but we should be able to have it run in O(N) be doing a single pass and replacing as we see something matching the conditions for replacement

pld (changes requested):

android/engine/src/main/java/org/smartregister/fhircore/engine/pdf/HtmlPopulator.kt
pld: stated simpler, there's an algorithm for this that only requires a single pass over the HTML. I'd be ok keeping the `while` loop style, if the searches are all from start to end of the string and you are truncating (not necessarily literally, you can use offsets) the string as you e.g., one you find a match for any `@` command, you can ignore any of the string before that match on any subsequent searches b/c you've already searched it.

Alternatively, and I think preferred, it seems reasonable to scan the string from front to back iterating overs its chars, when you see an `@` collect the chars until you hit a paren then see which it matches and do your substitution into an output string.

I'd guess what's going to be more performant depends on underlying lib optimizations, but either way the requirement is you only look at each character in the string a single time

pld (commented):
nice! this looks like it will be O(n)!

I don't see why there's a case where we do not increment i after going through the loop body, can you explain that?

pld (approved):

pld (approved):

PR #3322, opened 2024-06-07
IZ supported domain docs with built out interop section

pld:
This adds a documentation outline for the reference app related to immunization and builds out the content related to immunization.


docs/engineering/integrations/dhis2.mdx
pld, line 1: @ageryck I copied over evreything in the doc, did some small rewrites, except for the integration preq and assumptions parts, those seemed not as general

PR #3283, opened 2024-05-21
Fix navigation to Geowidget map on back press

aurangzaibumer:
* Fixed preventing map fragment from re-adding on the back stack if exists

* deleted unnecessary changes

**IMPORTANT: Where possible all PRs must be linked to a Github issue**

Fixes [link to issue]

**Engineer Checklist**
- [ ] I have written **Unit tests** for any new feature(s) and edge cases for bug fixes
- [ ] I have added any strings visible on UI components to the `strings.xml` file
- [ ] I have updated the  [CHANGELOG.md](./CHANGELOG.md) file for any notable changes to the codebase
- [ ] I have run `./gradlew spotlessApply` and `./gradlew spotlessCheck` to check my code follows the project's style guide
- [ ] I have built and run the FHIRCore app to verify my change fixes the issue and/or does not break the app 
- [ ] I have checked that this PR does NOT introduce **breaking changes** that require an update to **_Content_** and/or **_Configs_**? _If it does add a sample here or a link to exactly what changes need to be made to the content._

**Code Reviewer Checklist**
- [ ] I have verified **Unit tests** have been written for any new feature(s) and edge cases
- [ ] I have verified any strings visible on UI components are in the `strings.xml` file
- [ ] I have verifed the [CHANGELOG.md](./CHANGELOG.md) file has any notable changes to the codebase
- [ ] I have verified the solution has been implemented in a configurable and generic way for reuseable components
- [ ] I have built and run the FHIRCore app to verify the change fixes the issue and/or does not break the app

PR #3346, opened 2024-06-20
Fix Build | Refactor Fragment Backstack Transaction

ndegwamartin:
- Fix failing Unit Test
- Refactor NavController Backstack entry check due to Restricted API warning
- Add unit test for popBackStack if MainNavigationScreen.GeoWidgetLauncher.route route exists

### DIFF
--- a/android/quest/src/main/java/org/smartregister/fhircore/quest/util/extensions/ConfigExtensions.kt
+++ b/android/quest/src/main/java/org/smartregister/fhircore/quest/util/extensions/ConfigExtensions.kt
@@ -166,16 +166,14 @@ fun List<ActionConfig>.handleClickEvent(
       ApplicationWorkflow.DEVICE_TO_DEVICE_SYNC -> startP2PScreen(navController.context)
       ApplicationWorkflow.LAUNCH_MAP -> {
         val mapFragmentDestination = MainNavigationScreen.GeoWidgetLauncher.route
-        val isMapFragmentExists =
-          navController.currentBackStack.value.any { it.destination.id == mapFragmentDestination }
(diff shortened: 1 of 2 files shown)


android/quest/src/main/java/org/smartregister/fhircore/quest/util/extensions/ConfigExtensions.kt
```diff
@@ -166,16 +166,14 @@ fun List<ActionConfig>.handleClickEvent(
...
-        } else {
-          navController.navigate(
-            resId = mapFragmentDestination,
-            args = bundleOf(NavigationArg.GEO_WIDGET_ID to actionConfig.id),
-          )
         }
+        navController.navigate(
+          resId = mapFragmentDestination,
+          args = bundleOf(NavigationArg.GEO_WIDGET_ID to actionConfig.id),
+        )
```
ndegwamartin, line 176: **Note:** I've refactored the if else condition here such that now if the map fragment exists it is popped from the backstack and the controller navigates to the same(new) fragment. 

If this was not the intention then we can revert this part, let me know cc @dubdabasoduba @aurangzaibumer

ellykits: See my commit here https://github.com/opensrp/fhircore/pull/3283/commits/d703252255d0f27d29a7a8ce23b5cf5e16e93340/ 

This doesn't resolve the solution entirely though. It seems that it needs further investigation. Navigating from the profile back to the map exits the app instead of returning to the map's fragment.

ellykits:
I am closing this PR in favor of https://github.com/opensrp/fhircore/pull/3283.

Issue #3181, opened 2024-03-27
Volume Testing on the FHIR Core client app


ndegwamartin:
**Overview**
We need to conduct Volume Testing to see how the app scales with a lot of data.  This testing will help highlight the various bottlenecks with the app.

**Initial Test Data**
- 300 Households
- 10 Patients per house hold
- 30 Tasks per patient

We can gradually increase the data set size as we optimize and improve.

**Tasks**
- [x] Set up backend infrastructure
- [x] Set up user assignment data + Gateway configuration
- [x] Set up the dummy test data
- [x] Conduct volume testing and document results

Mstjamush:
**Device Specs**
```
Phone - Techno Spark 7
RAM - 3GB
ROM - 64GB
Android Version - 11
```
**_WiFi - 20Mbps
Number of Households - 1025_**

**Results**

- Initial Sync - 39:19.84
- Loading "Add Household" form -00:02.57
- Saving a Household - 0:45.06
- Loading "Add HH member" form - 00:02.38
- Saving a HH Member - 00:09.51
- Load HH profile - 00.11.68
- Load the All Households Register - 00:21.83

ellykits:
@Mstjamush Can you explicitly define the time format for the starts shared above? e.g. HH:MM:ss

Mstjamush:
@ellykits it's in HH:MM:ss format

pld:
@Mstjamush  can you share the same results for 10k households please?

Mstjamush:
@pld Let me generate the HHs and share

ndegwamartin:
**Observation on the Search Query:**
1. It uses the same Index to sort by as the register
2. Removing the index improves the performance as well
3. Search performance latency is directly proportional to the size of the resulting data set

**Opportunities for improvement**
1. Change search index to a non String type index e.g. last updated
2. Remove the index entirely - search narrows down the records enough
3. Limit triggering the db search until a minimum limit of characters is input by the user e.g. 3 chars

ndegwamartin:
Ticket to track performance optimizations on the FHIR SDK engine - https://github.com/google/android-fhir/issues/2547

ndegwamartin:
For the form loading, after profiling MWCore made the following observations for the Registration form:
1. First time load of the _Registration form_ is quite fast. Subsequent attempts take _12 to 13 seconds_
2. Loading the `Questionnaire` resource from the DB seems to introduce this latency

[...]

PR #3456, opened 2024-08-20
Batch searching when list might be large or when count is not defined

### LINKED ISSUE #3440: Reduce transaction(hog) time for GET that might return large results by fetching in batches

LZRS:

Fixes #3440 

**Engineer Checklist**

**Code Reviewer Checklist**


android/engine/src/main/java/org/smartregister/fhircore/engine/util/extension/FhirEngineExtension.kt
```diff
@@ -93,3 +96,29 @@ suspend fun FhirEngine.countUnSyncedResources() =
...
+    val pageCount = 100
+    do {
+      search.from = offset
+      search.count = pageCount
+      val searchResults = this.search<R>(search)
+      result += searchResults
+      offset += searchResults.size
+    } while (searchResults.size == pageCount)
+
+    result
```
ellykits, line 115: Thanks for this @LZRS. 

I have a few questions and comments on this implementation. 

The decision to load this data in batches sounds reasonable, however, based on this implementation I believe the gain would be negligible or there would be no gain at all. Technically, the data will still be cached in memory inside the loop. This differs from the approach where we would load and process the data in batches. Data processing should happen after fetching each batch to improve the app responsiveness, otherwise, there are some drawbacks with batching including:

1. Each batch fetch requires a new database query, introducing overhead. The overhead of executing smaller queries may outweigh the benefit of lower memory.

2. Using `OFFSET` requires the database to skip over records 
before returning the batch. A larger offset would require scanning and discarding many rows thus if indexing is not done properly, this will impact fetching data in batches. 

@LZRS Would it be possible to measure the performance impact introduced by this implementation for comparison?

The paging3 library loads data in batches and allows for data processing with each fetch. 

cc @ndegwamartin @pld

ellykits (changes requested):

android/engine/src/main/java/org/smartregister/fhircore/engine/util/extension/FhirEngineExtension.kt
LZRS: Yeah, generally, I think the performance gains would probably be negligible (in one of our cases, a difference of 20ms) especially when assuming only our query of interest would be running a single time. I think the significance of these changes would come about when we have multiple queries from different coroutines/threads.

Assuming for example we have a background long running operation accessing the db, and in the UI someone goes to a page that to render also requires a db fetch. Instead of waiting for the long running background operation to finish, the db fetch for the UI page can get queued in between the batches for the long running background. Looking to also add  integration tests on the same

The changes here would ensure that no single query(in batch) would take more than a couple of milliseconds, a long running query would still take long to be loaded (as it's split and cached in memory) but it would allow space for other queries to also get run.

I also agree that the best case would be data processing after fetching each batch although that would be quite tasking to implement, I feel.

In terms of MalawiCore, it improved the app's performance from [here](https://github.com/opensrp/fhircore/issues/3181#issuecomment-2134950554) to [here](https://github.com/opensrp/fhircore/issues/3181#issuecomment-2152451689).
We'll working on similar tests for FhirCore.

Use of paging3 to load would be the recommended approach but it would also need approval from the Google's sdk team

Note: 
> [Room](https://developer.android.com/reference/androidx/room/Transaction) will only perform at most one transaction at a time, additional transactions are queued and executed on a first come, first serve order.

ellykits (changes requested):

android/engine/src/androidTest/java/org/smartregister/fhircore/engine/util/extension/FhirEngineExtensionKtTest.kt
```diff
@@ -0,0 +1,85 @@
...
+        fhirEngine.batchedSearch<Patient> {}
+        println("Load patients..2")
+      }
+
+      launch {
+        fhirEngine.search<Questionnaire> {}
+        println("Load questionnaires..2")
+      }
+    }
+  }
```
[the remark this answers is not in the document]
LZRS: Done, updated the tests

ellykits (approved):

PR #3487, opened 2024-09-11
Add Exit Dialog

### LINKED ISSUE #3479: Override onBackPressed() and prompt user before exiting from App
**Desired Feature**
Override onBackPressed() and prompt the user before exiting from the App

1. onBackPressed() prompts the user if they want to Quit the application if on the Homescreen 
2. If in a given workflow, navigate through the back stack to the home screen then prompt the user to exit the app 
3. This should also be reproduced whether you use the physical back button or on the top bar navigation back button icon

**Acceptance Criteria** 

- [ ] For previously navigated screens traverse back the back stack list else if no home then exit the app after the prompt
- [ ] Make the on-back press text prompt translatable in the build variant string.xml file

hamza-vd:

Fixes #3479 

**Engineer Checklist**

**Code Reviewer Checklist**


android/quest/src/main/java/org/smartregister/fhircore/quest/ui/main/AppMainActivity.kt
```diff
@@ -298,4 +302,28 @@ open class AppMainActivity : BaseMultiLanguageActivity(), QuestionnaireHandler,
...
+            appMainViewModel.applicationConfiguration.appExitDialog?.message
+              ?: getString(R.string.exit_app_message)
+          AlertDialogue.showAlert(
+            this@AppMainActivity,
+            alertIntent = AlertIntent.CONFIRM,
+            title = title,
+            message = message,
+            cancellable = false,
+            confirmButtonListener = { finish() },
+            neutralButtonListener = { dialog -> dialog.dismiss() },
```
ellykits, line 323: @hamza-vd This alert should only be shown if there is no more screen to navigate back to. Can you confirm that is the case based on your implementation? I assume that every time the user clicks the back button then, they will be prompted via the dialog (according to my understanding of your code)

ellykits (changes requested):

hamza-vd:
Dialog content

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/main/AppMainActivity.kt
ellykits: This does thanks. @f-odhiambo ^

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/main/AppMainActivity.kt
```diff
@@ -300,4 +304,36 @@ open class AppMainActivity : BaseMultiLanguageActivity(), QuestionnaireHandler,
...
+        override fun handleOnBackPressed() {
+          val navController =
+            (supportFragmentManager.findFragmentById(R.id.nav_host) as NavHostFragment)
+              .navController
+          if (
+            navController.currentDestination?.id?.equals(R.id.geoWidgetLauncherFragment)!! ||
+              navController.currentDestination?.id?.equals(R.id.registerFragment)!!
+          ) {
+            val title =
+              appMainViewModel.applicationConfiguration.appExitDialog?.title
```
ndegwamartin, line 320: While it doesn't hurt to have it I don't think we need to make this configurable from App Configs. This is because the strings are already configurable via the **strings.xml** file and such a message is mostly likely only going to be configured when releasing an app and rarely after that. And since all apps should have it, it should not be optional (which would have justified the config)

I think we want to leave the app config level configurability for settings that can change frequently and most likely impact workflows which change regularly cc @ellykits

hamza-vd: @ellykits so what's the consensus on this? SHould I go ahead and remove the app config part?

ellykits: I missed that we can override the `strings` in the build variant. So yes, please we do not need the app configs now.

hamza-vd: Updated

ellykits (commented):
Looks good, see my comment.

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/main/AppMainActivity.kt
```diff
@@ -300,4 +304,30 @@ open class AppMainActivity : BaseMultiLanguageActivity(), QuestionnaireHandler,
...
+    onBackPressedDispatcher.addCallback(
+      object : OnBackPressedCallback(true) {
+        override fun handleOnBackPressed() {
+          val navController =
+            (supportFragmentManager.findFragmentById(R.id.nav_host) as NavHostFragment)
+              .navController
+          if (
+            navController.currentDestination?.id?.equals(R.id.geoWidgetLauncherFragment)!! ||
+              navController.currentDestination?.id?.equals(R.id.registerFragment)!!
+          ) {
```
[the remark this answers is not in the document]
hamza-vd: Condition updated

ellykits (approved):

PR #3528, opened 2024-10-01
Add `nothing` app workflow

FikriMilano:

Change:
Add `NOTHING` as new ApplicationWorkflow.

Why:
Useful when the ApplicationWorkflow is configurable, when there's certain use case where if the user clicks a button, it does nothing.

**Engineer Checklist**

**Code Reviewer Checklist**


ellykits:
I do not understand the use case for this @FikriMilano. How will this be used?

FikriMilano:
say @ellykits can we simply declare a `null` in rules? will that make the button does nothing when clicked?
```
{
  "name": "onClick",
  "condition": "true",
  "priority": 2,
  "actions": [
    "data.put('onClick', data.get('taskCode').equals('029') ? 'LAUNCH_PROFILE' : null)"
  ]
}
```

FikriMilano:
Based on this line, it should be ok to declare a null
https://github.com/opensrp/fhircore/blob/main/android/quest/src/main/java/org/smartregister/fhircore/quest/util/extensions/ConfigExtensions.kt#L103

ellykits:
> Say @ellykits can we simply declare a `null` in rules? will that make the button does nothing when clicked? This new workflow was meant to replace `null` value as I thought we can't declare one while inside a rule.
> 
> ```
> {
>   "name": "onClick",
>   "condition": "true",
>   "priority": 2,
>   "actions": [
>     "data.put('onClick', data.get('taskCode').equals('029') ? 'LAUNCH_PROFILE' : null)"
>   ]
> }
> ```

You can use the rule condition instead. In your case, the rule should only be executed if the task code is '029'. Re-write the rule to: 

```json
{
  "name": "onclickAction",
  "condition": "data.get('taskCode').equals('029')",
  "priority": 2,
  "actions": [
    "data.put('onclickAction', 'LAUNCH_PROFILE')"
  ]
}
```

ellykits:
Alternatively, you can use the `clickable` attribute that accepts a boolean from an executed rule. @FikriMilano. If the check is not performed before a click action for your widget, could you implement it? Theoretically, it should work for all supported widgets.

ellykits:
I'm closing this, use the suggestions in the comments. Feel free to re-open this if the suggestions do not work @FikriMilano.

Issue #3565, opened 2024-10-18
Update the GeoWidget Topbar to add the filter Icon next the Navigate to register one


dubdabasoduba:
**Describe the feature request.**
- We would like to update the GeoWidget module to add a filter button next to the navigate to register one.  
- This button should have the following properties
   - Be configurable i.e can be hidden and shown by changing configs
   - Can launch questionnaires
   - Can launch the multi-select widget 

**Additional context**
- Due to the performance issues we are facing on the maps we would like to decouple the filtering of data on the maps from the multi-select widget used to determine what data to sync.  
- We would then move the filtering to a multi-select launched from the filtering icon described above. 

**Acceptance criteria**
- [ ] Be able to show a filter icon on the geowidget toolbar
- [ ] This icon should be configurable as defined above
- [ ] This widget should be able to launch the workflows defined above

**Implementation plan (For Engineers)**
The plan for implementing the solution e.g. via a _description_ or a _check list_ for the various ordered tasks that will need to be completed.
_i.e. Describe how you intend to solve the problem_

AnnieMungai:
EUSM v.2.0.0.RC12
LGTM!

Issue #3567, opened 2024-10-18
Decouple sync and register/map filtering control from the multiselect widget


dubdabasoduba:
**Describe the feature request.**
- The current multi-select widget helps in downloading and filtering data.  We would like to decouple these two functions but leave the same functionality on the multi-select widget
- We can add configuration to decide what function the multi-select widget should take.  

**Acceptance criteria**
- [ ] Decouple the sync and filtering functionality on the multi-select widget.  
- [ ] Selection of the functionality should be determined by configurations.  

**Implementation plan (For Engineers)**
The plan for implementing the solution e.g. via a _description_ or a _checklist_ for the various ordered tasks that will need to be completed.
_i.e. Describe how you intend to solve the problem_

AnnieMungai:
EUSM v2.0.0RC12
LGTM!

Issue #3566, opened 2024-10-18
Add a select one mode to the multiselect widget


dubdabasoduba:
**Describe the feature request.**
- Due to the performance issues we are facing on the maps we would like to add a select one mode to the multi-select widget. 
- This means the users will only be allowed to select one item at a time.  If a second item is selected then the initially selected one will be deselected. 
- This will be used on the filter defined on this icon https://github.com/opensrp/fhircore/issues/3565 

**Acceptance criteria**
- [ ] Add a select one mode that restricts the users to only select one root at a time.  
- [ ] If another root is selected then the previous one is deselected. 

**Implementation plan (For Engineers)**
The plan for implementing the solution e.g. via a _description_ or a _checklist_ for the various ordered tasks that will need to be completed.
_i.e. Describe how you intend to solve the problem_

dubdabasoduba:
@ndegwamartin I am closing this since it was completed

PR #3568, opened 2024-10-18
Decouple multi select data sync and filter

ellykits:

Fixes https://github.com/opensrp/fhircore/issues/3565, https://github.com/opensrp/fhircore/issues/3566, https://github.com/opensrp/fhircore/issues/3567

**Engineer Checklist**

**Code Reviewer Checklist**


ellykits:
I introduced two new config properties:

```json
 "viewActions": [
     "FILTER_DATA", 
     "SYNC_DATA"
  ]
```
and 
```json
"mutuallyExclusive": true
```

These two properties (`viewActions` and `mutuallyExclusive`) customize the actions and behavior of the multi-select choice checkbox, respectively.

android/geowidget/src/main/java/org/smartregister/fhircore/geowidget/screens/GeoWidgetFragment.kt
```diff
@@ -311,33 +306,44 @@ class GeoWidgetFragment : Fragment() {
...
+        val locationPoints =
+          featureCollection
+            .features()
+            ?.asSequence()
+            ?.filter { it.geometry() is Point }
+            ?.map { it.geometry() as Point }
+            ?.toMutableList() ?: emptyList()
+
+        val bbox = TurfMeasurement.bbox(MultiPoint.fromLngLats(locationPoints))
+        val paddedBbox = CoordinateUtils.getPaddedBbox(bbox, 1000.0)
```
pld, line 330: What about moving 1000 to a constant so we can describe it better?

android/geowidget/src/main/java/org/smartregister/fhircore/geowidget/screens/GeoWidgetFragment.kt
```diff
@@ -311,33 +306,44 @@ class GeoWidgetFragment : Fragment() {
...
+        val bounds = LatLngBounds.from(paddedBbox[3], paddedBbox[2], paddedBbox[1], paddedBbox[0])
```
pld, line 331: I don't think so, but does kotlin have a destructure into args syntax? I think in Python we could do *paddedBox[:3].reverse() instead of adding all the indices

pld (approved):

android/geowidget/src/main/java/org/smartregister/fhircore/geowidget/screens/GeoWidgetFragment.kt
ellykits: Yea, refactoring this to a more Kotlinish code will make this more readable. Lemme check. This implementation borrowed from Kujaku Java code.

ellykits: Kotlin has a reversed function.

ellykits: Oh! we can't update this to use List, the function is overloaded, and the values are function params.

pld (approved):

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/main/components/TopScreenSection.kt
```diff
@@ -134,62 +143,68 @@ fun TopScreenSection(
...
+              imageVector = Icons.Outlined.FilterAlt,
+              contentDescription = FILTER,
+              tint = Color.White,
+              modifier =
+                modifier
+                  .clickable { onClick(ToolbarClickEvent.FilterData) }
+                  .testTag(TOP_ROW_FILTER_ICON_TEST_TAG),
+            )
+          }
         }
```
dubdabasoduba, line 199: @ellykits Is there a reason for rearranging the icon placement?  This makes the filter icon to always default as first on registers

dubdabasoduba (changes requested):
@ellykits Great! job on this refactor. I have tested it out and I think we have a few bugs.  

- The 2 instances of the multiselect seem to share the same `sharedPRef` entry i.e if I select locations on the multiselect meant for sync, the same locations would be selected on the multiselect meant for filtering.  
- For some reason the toasters and register messages meant to notify the user when no data is fetched don't seem to be displayed if the filter should return 0 results.  

Here is a video of the bugs
https://github.com/user-attachments/assets/d4d7a529-4fe5-4fc9-b977-6a0cea9ce442

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/main/components/TopScreenSection.kt
ellykits: Yes, this is because I introduced an overflow menu at the end if there are more than two icons. This register filter implementation was implemented as a standalone, it's not activated via the `topScreenMenuIcons` actions.

ellykits: Can I suggest we use a different icon for the map filter to avoid this confusion?

ellykits: @HenryRae Can you help with this? Can we have an icon to represent filtering data on a map for instance placing the funnel icon on top of the map icon?

pld: can you share a screenshot? it's not clear to me why we'd want a different icon, since the action (filtering) is the same, and that's what we want the icon to represent

pld: @dubdabasoduba showed me a screen, he'll reach out to clarify

PR #3647, opened 2024-12-09
Update Composition.focus to handle Composition.entry

### LINKED ISSUE #3617: Update OpenSRP to download resources referenced on the `Composition.focus` and `Composition.entry` attributes
- The OpenSRP mobile app uses the Composition resource to define the resources that need to be downloaded to the app for configurations and workflow execution.
    - For a composition to be validated, every section must include at least one of the following: an `entry`, `text`, or `sub-section`.
- This ticket involves updating the OpenSRP mobile app code to support both the entry and focus attributes for referencing resources the app requires. This update will maintain backward compatibility with existing implementations.
- `focus` should be deprecated over time, and users will be encouraged to transition to the `entry` attribute for resource referencing.

**Additional context**
- Teams are advised to use the `text` property for Composition resources that haven't migrated to using `entry` to pass validation.   Below is an example of how to use `text` 
```json
{
  "title": "Application configuration details",
  "text": {
    "status": "generated",
    "div": "<div xmlns=\"http://www.w3.org/1999/xhtml\">Application configuration details</div>"
  },
  "focus": {
    "reference": "Binary/592f27bc-d476-4d8f-9307-07e00f1b91bb",
    "identifier": {
      "value": "application"
    }
  },
  "mode": "working"
}
```

**Acceptance criteria**
- [x] Composition resources defined using the `entry` attribute should load correctly on the app. 
- [ ] All resources referenced on the `entry` and `focus` attributes should be downloaded correctly.  

**Area path**
A list of ordered steps in the app on the usage of the feature to support anyone testing it e.g. Code reviewer, QA e.g.
1. Login to the app
2. Open _Navigation bar_
3. Click on _Children register_
4. Click on _Child profile_
5. Click on _Edit profile_ from menu

The plan for implementing the solution e.g. via a _description_ or a _checklist_ for the various ordered tasks that will need to be completed.

sharon2719:

Fixes #3617 

**Engineer Checklist**

**Code Reviewer Checklist**


android/engine/src/main/java/org/smartregister/fhircore/engine/configuration/ConfigurationRegistry.kt
pld, line 1: Can you rewrite this by defining a function that takes the values that have the key of either "focus" or "entry", so that you can keep all the code the same other than where you happen to get the values from?

E.g. regardless of whether the key in the composition is "focus" or "entry" all the process that happens to the 
```
{
    "reference": ...
   "identifier": ...
}
```
is going to be exactly the same. So the only part that should vary is whether you get that from the a key that's called "focus" or "entry".

Also, not that in the test cases you need to handle at least the cases where

1. the composition uses all focus
2. the composition uses all entry
3. the composition uses both, but not in the same section
4. the composition uses both and at least one in the same section (ie verify it always takes either entry or focus when it sees both)
5. the composition has neither entry or focus (yes, this is invalid but we need to check it doesn't crash)

pld: Great start!

sharon2719:
Closed this in favor of this #https://github.com/opensrp/fhircore/pull/3662

PR #3704, opened 2025-01-23
Add Fetch Resource by Reference in RulesEngine

qiarie:

Fixes [link to issue]

**Engineer Checklist**

**Code Reviewer Checklist**

### DIFF
--- a/android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
+++ b/android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
@@ -33,8 +33,6 @@ import java.util.Locale
 import javax.inject.Inject
 import javax.inject.Singleton
 import kotlin.system.measureTimeMillis
-import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.launch
 import kotlinx.coroutines.runBlocking
 import org.apache.commons.jexl3.JexlEngine
 import org.hl7.fhir.r4.model.Base
--- a/android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
+++ b/android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
@@ -19,6 +19,7 @@ package org.smartregister.fhircore.engine.rulesengine
 import android.content.Context
 import ca.uhn.fhir.context.FhirContext
 import com.google.android.fhir.datacapture.extensions.logicalId
+import com.google.android.fhir.db.ResourceNotFoundException
 import com.google.android.fhir.search.Order
 import com.jayway.jsonpath.Configuration
 import com.jayway.jsonpath.JsonPath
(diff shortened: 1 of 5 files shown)


android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
```diff
@@ -774,6 +777,29 @@ constructor(
...
+      }
+    }
+
+    fun getResourceByResourceId(
+      resourceType: String,
+      resourceId: String,
+    ): Resource? {
+      return runBlocking(dispatcherProvider.io()) {
+        runCatching {
+            defaultRepository.loadResource(resourceId, ResourceType.valueOf(resourceType))
```
ellykits, line 798: Can you add tests for these

ellykits (changes requested):

android/quest/src/main/java/org/smartregister/fhircore/quest/data/DataMigration.kt
```diff
@@ -199,23 +201,26 @@ constructor(
...
                     value,
                   )
+                  baseResourceUpdated = true
                 }
               }
             }
 
-          val resourceDefinition: Class<out IBaseResource>? =
-            FhirContext.forR4Cached().getResourceDefinition(resource).implementingClass
+          if (baseResourceUpdated) {
```
Rkareko, line 209: Add / update tests for the `baseResourceUpdated` `true` and `false` scenarios

pld (dismissed):

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
```diff
@@ -774,6 +778,39 @@ constructor(
...
     }
+
+    fun getResourceByReference(
+      resourceReference: String?,
+    ): Resource? {
+      if (resourceReference.isNullOrEmpty()) {
+        return null
+      }
+
+      return runBlocking(dispatcherProvider.io()) {
```
ellykits, line 789: If you call runBlocking(Dispatchers.IO) from the main thread, the main thread will be blocked, but the coroutine itself will run on a thread from the Dispatchers.IO pool. Use `CoroutineScope` instead.

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
```diff
@@ -774,6 +778,39 @@ constructor(
...
+
+    fun getResourceByIdAndType(
+      resourceId: String?,
+      resourceType: String?,
+    ): Resource? {
+      if (resourceId.isNullOrEmpty() || resourceType.isNullOrEmpty()) {
+        return null
+      }
+
+      return runBlocking(dispatcherProvider.io()) {
```
ellykits, line 806: This will block the main thread where the rules are executed. Use a new CoroutineScope instead.

ellykits (changes requested):

android/engine/src/main/java/org/smartregister/fhircore/engine/rulesengine/RulesFactory.kt
qiarie: Isn't `CoroutineScope` ideal when we don't need to return a value? CoroutineScope may return earlier before we get return value.

ellykits: Not really, CoroutineScope. You can return using `async` function of the CoroutineScope. 

i.e.
```kotlin 
val coroutineScope = CoroutineScope(dispatcherProvider.io())

val result = coroutineScope.async {
// Your logic that returns a value
}.await()

```
`async` will defer the returned value when the `await` function is called.

PR #3738, opened 2025-03-31
🐛 Remove `#` when saving practitioner details

dubdabasoduba:

Fixes [link to issue]

**Engineer Checklist**

**Code Reviewer Checklist**

### DIFF
--- a/android/quest/src/main/java/org/smartregister/fhircore/quest/ui/login/LoginViewModel.kt
+++ b/android/quest/src/main/java/org/smartregister/fhircore/quest/ui/login/LoginViewModel.kt
@@ -30,8 +30,13 @@ import androidx.work.WorkManager
 import dagger.hilt.android.lifecycle.HiltViewModel
 import io.sentry.Sentry
 import io.sentry.protocol.User
+import java.net.SocketTimeoutException
+import java.net.UnknownHostException
+import java.util.concurrent.TimeUnit
+import javax.inject.Inject
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
+import org.hl7.fhir.r4.model.Bundle as FhirR4ModelBundle
 import org.hl7.fhir.r4.model.CareTeam
 import org.hl7.fhir.r4.model.Group
 import org.hl7.fhir.r4.model.Location
@@ -69,11 +74,6 @@ import org.smartregister.model.location.LocationHierarchy
 import org.smartregister.model.practitioner.PractitionerDetails
 import retrofit2.HttpException
 import timber.log.Timber
-import java.net.SocketTimeoutException
-import java.net.UnknownHostException
-import java.util.concurrent.TimeUnit
-import javax.inject.Inject
-import org.hl7.fhir.r4.model.Bundle as FhirR4ModelBundle
 
 @HiltViewModel
 class LoginViewModel
--- a/android/quest/src/main/java/org/smartregister/fhircore/quest/ui/login/LoginViewModel.kt
+++ b/android/quest/src/main/java/org/smartregister/fhircore/quest/ui/login/LoginViewModel.kt
@@ -30,13 +30,8 @@ import androidx.work.WorkManager
 import dagger.hilt.android.lifecycle.HiltViewModel
 import io.sentry.Sentry
 import io.sentry.protocol.User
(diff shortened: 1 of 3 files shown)


Rkareko (dismissed):

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/login/LoginViewModel.kt
```diff
@@ -365,25 +367,25 @@ constructor(
           containedResources.forEach { resource ->
             when (resource.resourceType) {
               ResourceType.CareTeam -> {
-                careTeams.add(resource as CareTeam)
+                careTeams.add(updateResourceIds(resource) as CareTeam)
```
allan-on, line 370: Since `updateResourceIds(resource)` is called for every resource type (in the when statement) and isn't dependent on the type, would it be better to call that function once before the `when` statement?

android/quest/src/main/java/org/smartregister/fhircore/quest/util/Utils.kt
```diff
@@ -0,0 +1,37 @@
...
+/** Utility object containing helper functions for common operations. */
+object Utils {
+  /**
+   * Removes any '#' characters that appear before the first string or integer value.
+   *
+   * @param resourceId The string or integer value that may have multiple '#' characters at the
+   *   start
+   * @return The value without any '#' characters at the start
+   */
+  fun removeHashPrefix(resourceId: Any): String {
```
allan-on, line 28: Are all logical IDs for all resources expected not to have the `#`, or is this only for the metadata resources being modified here?

allan-on (commented):
See coments

android/quest/src/main/java/org/smartregister/fhircore/quest/util/Utils.kt
dubdabasoduba: All logical Ids for resources added to the `contained` are expected to have `#` when returned by the server. The idea here is to update the resources returned by the PractitionerDetail endpoint to remove the `#`. 

We might need to do this in other places if we are using resources in the `contained` property

android/quest/src/main/java/org/smartregister/fhircore/quest/util/Utils.kt
```diff
@@ -0,0 +1,37 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.smartregister.fhircore.quest.util
+
+/** Utility object containing helper functions for common operations. */
+object Utils {
```
ellynerdstone, line 20: This utility can be refactored to an extension

android/quest/src/main/java/org/smartregister/fhircore/quest/util/Utils.kt
ellynerdstone, line 28: How about we update the function for `extractLogicalUuid` only to trim this prefix?

android/quest/src/main/java/org/smartregister/fhircore/quest/util/Utils.kt
ellynerdstone: Re-use string extension file

ellynerdstone (changes requested):

android/quest/src/main/java/org/smartregister/fhircore/quest/util/Utils.kt
dubdabasoduba: mmmh, while this makes sense it assumes we what to remove the # everywhere. The hash is needed for resources that actually use contained. 


We are removing them on the practitioner details because we don't use the details while in the contained.

android/quest/src/main/java/org/smartregister/fhircore/quest/util/Utils.kt
dubdabasoduba: I can move it there

dubdabasoduba: Moved

ellykits (approved):

PR #3843, opened 2026-06-11
Prevent logout while filling in a questionnaire (+ offline redirect fix)

### LINKED ISSUE #3842: Offline session-expired re-authentication loop after questionnaire submit
## Context 

When a user's refresh token is expired and the device is offline, every questionnaire submission shows a "session expired" dialog and routes to the login screen. 

Image

Offline login succeeds, returning them to the app; the next submission triggers the dialog again, and so on. 
The user is unable to exit the loop while offline because the only way to resolve the condition is to issue a new refresh token, which requires network connectivity. 

### Check for connectivity before showing the dialog. 

- Offline → no dialog; the user keeps working and saving locally. 
- The expired session then surfaces naturally when connectivity returns and a sync / token-requiring call reaches TokenAuthenticator.getAccessToken(), which performs re-authentication. 

```java
// AppMainActivity.startForResult
if (this@AppMainActivity.isDeviceOnline() &&
    !tokenAuthenticator.isCurrentRefreshTokenActive()) {
  delay(3000)
  showSessionExpiredDialog()
}
```

#### Removing the dialog 
Deleting the block at AppMainActivity.kt:127-132 does stop the loop (the loop is driven entirely by that dialog → LoginActivity navigation), but it doesn't remove the expired-session problem. It shifts re-authentication onto the implicit fallback path in `TokenAuthenticator.getAccessToken()` (TokenAuthenticator.kt:73), 

- Offline: user submits, nothing happens, keeps working. The user isn't informed. 
- Online (expired refresh token): no "form saved" prompt. User keeps working until the next token-requiring call (a periodic or manual sync) hits getAccessToken(), fails to refresh, and the Login Activity is launched.

allan-on:
Brings the *"prevent redirecting to the login screen in the middle of filling in a Questionnaire"* implementation to `main`, together with its follow-up bug fix.

Both changes currently live only on `custom-engage`; this PR cherry-picks them onto a branch off `main`:
1. **`Prevent logout when filling questionnaire`** (cherry-picked from `75eb3468d`, by @fikrimilano)
2. **`If refresh token expired redirect to login if online`** (cherry-picked from `81240e7d5`) — fixes #3842

### DIFF
--- a/android/engine/src/main/java/org/smartregister/fhircore/engine/data/remote/shared/TokenAuthenticator.kt
+++ b/android/engine/src/main/java/org/smartregister/fhircore/engine/data/remote/shared/TokenAuthenticator.kt
@@ -67,6 +67,9 @@ constructor(
   private val authConfiguration by lazy { configService.provideAuthConfiguration() }
   private var isLoginPageRendered = false
 
+  /** When true, suppresses automatic LoginActivity launch (e.g., during questionnaire filling) */
+  var suppressLoginRedirect = false
+
   fun getAccessToken(): String {
     val account = findAccount() ?: return ""
     val accessToken = accountManager.peekAuthToken(account, AUTH_TOKEN_TYPE) ?: ""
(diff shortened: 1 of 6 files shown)


LZRS (approved):

ellykits (changes requested):
Thanks for the PR @allan-on, while this temporarily fixes the issue, the underlying reason why this happen is not actually addressed in this PR. What we are doing is suppressing the redirect to the login page, via a class property flag. What would happen when the class is garbage collected? Can we think about resolving this in  a different way?

allan-on:
Changes applied in https://github.com/opensrp/fhircore/pull/3847

PR #3844, opened 2026-06-11
Remove lingering sync overlay scrim on low-resource devices

### LINKED ISSUE #3841: Blocking Sync progress state remains permanent in some use cases
**Describe the bug**

- Application remains at 0% sync state when user navigates out of the applicaiton during sync process
- Application remains in blocking state perpetually after sync gets to 100%

**Screenshots**

Image
Image

 - Device: Tecno Spark 7 (Model No: TECNO KF6j)
 - Android version: 11
 - Build number: KF6j-H6218A-R-GL-23011V640
 - App Version [e.g. 21(2.2.3)]

qiarie:

Fixes #3841

**Engineer Checklist**

**Code Reviewer Checklist**

### DIFF
--- a/android/engine/src/main/java/org/smartregister/fhircore/engine/ui/components/register/LoaderDialog.kt
+++ b/android/engine/src/main/java/org/smartregister/fhircore/engine/ui/components/register/LoaderDialog.kt
@@ -109,11 +109,6 @@ private fun LoaderContent(
   showBackground: Boolean,
   showLineSpinIndicator: Boolean,
 ) {
-  // NOTE: This composable intentionally has no self-dismiss logic. Its presence in the
-  // composition (and therefore the surrounding Dialog window + scrim) is controlled solely by the
-  // caller. A previous SideEffect hid this content at 100% while leaving the Dialog window - and
-  // its full-screen scrim - composed, which orphaned a touch-blocking overlay on screen. Content
-  // and window must appear and disappear together; the caller is the single source of truth.
   Box(modifier.size(boxWidth, boxHeight)) {
     Column(
       modifier = modifier.padding(8.dp),
--- a/android/engine/src/main/java/org/smartregister/fhircore/engine/ui/components/register/LoaderDialog.kt
+++ b/android/engine/src/main/java/org/smartregister/fhircore/engine/ui/components/register/LoaderDialog.kt
@@ -27,10 +27,7 @@ import androidx.compose.material.CircularProgressIndicator
 import androidx.compose.material.Surface
 import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.SideEffect
 import androidx.compose.runtime.collectAsState
-import androidx.compose.runtime.mutableStateOf
-import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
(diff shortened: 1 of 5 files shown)


qiarie:
# First-time sync dialog: lingering overlay/scrim fix

## Symptom

After the initial data sync reaches 100%, the progress dialog appears dismissed, but on some devices (typically lower-resource ones) the dim **overlay/scrim** that accompanies the modal dialog remains on screen and blocks all touch input - the app becomes unusable. The stuck overlay can appear to persist across app close/reopen.

## Root cause

The "overlay" is the **scrim (dim-behind window) of a Compose `Dialog`**. The dialog had **two independent, unsynchronized dismiss mechanisms** that targeted different things and fired at different times.

The dialog is built in two nested layers (`engine/.../register/LoaderDialog.kt`):

- The **window** (and its scrim) is the Compose `Dialog(...)` - present whenever `LoaderDialog` is
  in the composition.
- The **content** (spinner + "NN%" box) is `LoaderContent`, which carried its own local
  `remember { mutableStateOf(true) }` and self-hid via a `SideEffect`:

  ```kotlin
  SideEffect {
    if (currentPercentage >= 100) {
      openDialog.value = false   // hides CONTENT only - NOT the Dialog window/scrim
    }
  }
  ```

What actually controlled the scrim was the caller (`RegisterScreen.kt`):

```kotlin
if (!BuildConfig.SKIP_AUTHENTICATION && registerUiState.isFirstTimeSync) {
  LoaderDialog(...)   // only this removes the whole Dialog window + scrim
}
```

So the two mechanisms fired on different signals:

| Mechanism | Hides what | Fires on |
|---|---|---|
| `SideEffect` at `>= 100` | inner content box only | progress hits 100% - during the `Running`/`InProgress` phase |
| `isFirstTimeSync → false` | the whole Dialog **+ scrim** | terminal `CurrentSyncJobStatus.Succeeded` → multi-step recompute |

`calculatePercentageProgress` (`AppMainViewModel.kt`) reaches 100 from `completed == total` **while sync is still `Running`** - strictly before `Succeeded` is emitted. The resulting sequence:

1. Progress hits 100% → `SideEffect` hides the content. The user perceives "dialog dismissed."
2. The `Dialog` window is **still composed**, so its full-screen scrim remains - now an empty, touch-blocking dim layer.
3. Only later, when `Succeeded` arrives and `isFirstTimeSync` recomputes to `false`, is the scrim torn down.

On a fast device step 2→3 is milliseconds and the orphaned scrim is never noticed. On a low-resource device the gap is long and visible - and if step 3 never completes, the scrim is **permanent**.

### Why step 3 was fragile (and why it "persists across open/close")

Flipping `isFirstTimeSync` to false was a derived value recomputed through a fragile chain:

```kotlin
// RegisterViewModel.kt
private fun isFirstTimeSync() =
  read(LAST_SYNC_TIMESTAMP).isNullOrEmpty() &&
    applicationConfiguration.usePractitionerAssignedLocationOnSync &&
    _totalRecordsCount.longValue == 0L
```

Three fragilities:

1. **Two listeners, no ordering guarantee.** The `LAST_SYNC_TIMESTAMP` write happens in
   `AppMainViewModel` (via `AppMainActivity.onSync`), but the `isFirstTimeSync()` recompute happens in a *separate* listener, `RegisterFragment.onSync → refreshRegisterData`. If the fragment recomputes before the timestamp is written, the flag stays `true`.
2. **Stale count race.** In `retrieveRegisterUiState`, `registerUiState.value` was assigned **synchronously** with `isFirstTimeSync()` while `_totalRecordsCount` was updated inside a *separately launched* coroutine - so the flag read the previous (stale) count, still `0`.
3. **Terminal-state coupled to data, not to "sync done."** Dismissal depended on `_totalRecordsCount > 0` and a written timestamp. A failed/partial sync (no timestamp written), or a register whose count stays `0` due to related-entity-location filtering, leaves `isFirstTimeSync()` returning `true` on every launch → the dialog re-shows, hits 100%, content hides, scrim orphaned again. That is the "persists between opening and closing."

The `SideEffect` is the *direct* cause of the orphaned scrim; the derived-state dismissal chain is *why it gets stuck* on slow/unlucky runs.

## Fix summary

Make the scrim and its dismissal driven by **one authoritative signal**, and remove the half-measure that hid content independently of the window.

### #1 - `engine/.../register/LoaderDialog.kt` (orphaned-scrim root cause)

Removed `LoaderContent`'s local `openDialog` state and the`SideEffect { if (currentPercentage >= 100) openDialog.value = false }`. Content and window now live and die together, gated entirely by the caller. Dropped the now-unused `SideEffect`, `remember`, and `mutableStateOf` imports.

### #2 - `quest/.../register/RegisterScreen.kt` (reliable teardown signal)

The dialog is now gated on `isFirstTimeSync && !syncTerminated`, where:

```kotlin
val syncTerminated =
  appDrawerUIState.currentSyncJobStatus is CurrentSyncJobStatus.Succeeded ||
    appDrawerUIState.currentSyncJobStatus is CurrentSyncJobStatus.Failed
```

This is delivered to all listeners on the same event, so teardown no longer waits on the fragile async `isFirstTimeSync` recompute. It also clears on a **failed** sync rather than trapping the user. Added the `CurrentSyncJobStatus` import.

### #3 - `quest/.../register/RegisterViewModel.kt` (stale-count race)

`retrieveRegisterUiState` built `RegisterUiState(isFirstTimeSync = isFirstTimeSync(), …)` synchronously while `_totalRecordsCount` was updated in a separate coroutine. Extracted a `buildRegisterUiState()` helper and re-emit it inside the count coroutine once `_totalRecordsCount` is fresh. The dialog still shows immediately on first launch (synchronous emit), then dismisses once the count reflects synced data.

## Note

The dialog now also clears on **`Failed`** (chosen so a failed sync does not strand the user behind a scrim). If the dialog should instead stay until a sync actually succeeds, drop the `Failed` branch from `syncTerminated` in `RegisterScreen.kt`.

android/engine/src/main/java/org/smartregister/fhircore/engine/ui/components/register/LoaderDialog.kt
```diff
@@ -112,65 +109,62 @@ private fun LoaderContent(
...
-    Box(modifier.size(boxWidth, boxHeight)) {
-      Column(
-        modifier = modifier.padding(8.dp),
-        verticalArrangement = Arrangement.Center,
-        horizontalAlignment = Alignment.CenterHorizontally,
+  // NOTE: This composable intentionally has no self-dismiss logic. Its presence in the
+  // composition (and therefore the surrounding Dialog window + scrim) is controlled solely by the
+  // caller. A previous SideEffect hid this content at 100% while leaving the Dialog window - and
+  // its full-screen scrim - composed, which orphaned a touch-blocking overlay on screen. Content
+  // and window must appear and disappear together; the caller is the single source of truth.
```
ellykits: Delete these coments

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/register/RegisterViewModel.kt
```diff
@@ -666,20 +679,17 @@ constructor(
                   )
                   .toInt(),
             )
+
+          // Re-emit with isFirstTimeSync recomputed now that the up-to-date record count is
+          // available. The synchronous assignment below runs before this coroutine resolves, so it
+          // reads a stale (previous) _totalRecordsCount; without this re-emission a register that
+          // already holds synced data could keep isFirstTimeSync == true and strand the sync
+          // dialog.
```
ellykits: Delete these comments

ellykits (changes requested):

ellykits:
@qiarie Whats the feedback after testing this on various apps? an you share the details in a comment?

android/quest/build.gradle.kts
```diff
@@ -440,6 +440,16 @@ android {
       "app_name",
       "\"${variant.mergedFlavor.manifestPlaceholders["appLabel"]}\"",
     )
+
+    variant.outputs.all {
+      val output = this as com.android.build.gradle.internal.api.BaseVariantOutputImpl
```
ellykits: Import `BaseVariantOutputImpl` instead of using qualified import here.

ellykits (changes requested):

ellykits:
> @qiarie Whats the feedback after testing this on various apps? an you share the details in a comment?

I like that we are no longer using side effect to react to the sync progress.

ellynerdstone:
> > @qiarie Whats the feedback after testing this on various apps? an you share the details in a comment?
> 
> Ignore the graphics issue, it's an issue with the phone

Which graphics issue? the circular progress bar animation?

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/register/RegisterScreen.kt
```diff
@@ -172,7 +173,15 @@ fun RegisterScreen(
     },
   ) { innerPadding ->
     Box(modifier = modifier.padding(innerPadding)) {
-      if (!BuildConfig.SKIP_AUTHENTICATION && registerUiState.isFirstTimeSync) {
+      // Tear down the first-time sync dialog (and its full-screen scrim) on a single authoritative
+      // signal: the sync reaching a terminal state. Relying on isFirstTimeSync alone left the scrim
+      // composed until a fragile, async recompute (timestamp write + register count) landed, which
+      // could lag badly or never complete on low-resource devices, stranding a touch-blocking
+      // overlay. The terminal CurrentSyncJobStatus is delivered to all listeners on the same event.
```
ellynerdstone, line 180: These comments are really not necessary,  when you reading the code below

ellynerdstone (dismissed):

qiarie:
Yes. It was an issue with my phone

android/quest/src/main/java/org/smartregister/fhircore/quest/ui/register/RegisterScreen.kt
ellykits: Delete the inline comments. Otherwise this looks good.

ellykits (approved):

